PR #3, opened 2021-12-28
Add binary protocol/prepared statement support

brianmario:
As the title says, this adds support for the binary protocol, which is used by prepared statements.

There's still a little bit to do here, though mainly just wrapping up docs and some cleanup and adding a few more tests. But I wanted to get this open for review sooner than later.

I figured wrapping this with the Ruby gem/extension could happen in a separate pull request, but if it makes sense to just go ahead and add here let me know.

Some of the things I'd like to wrap up before calling this done:

- [x] Finish up writing client API docs
- [x] Finish up writing blocking API docs
- [x] Add tests for the blocking API
- [x] Fix null bitmap builder
- [x] Add more tests for edge cases
- [x] Document `trilogy_binary_value_t`, describing which C types map to which MySQL column types.

So happy to see this project finally made open source.

Thank you!

### DIFF
--- a/inc/trilogy/protocol.h
+++ b/inc/trilogy/protocol.h
@@ -621,39 +621,125 @@ typedef struct {
     size_t data_len;
 } trilogy_value_t;
 
+/* trilogy_binary_value_t - MySQL binary protocol value type
+ *
+ * See https://dev.mysql.com/doc/internals/en/binary-protocol-value.html for more detail.
+ */
 typedef struct {
+    // Flag denoting the value is NULL.
     bool is_null;
+
+    /* Flag denoting the numeric value is unsigned.
+     * If this is true, the unsigned numerical value types should be used
+     * from the `as` union below.
+     *
+     * For example, if the value's MySQL type is TRILOGY_TYPE_LONGLONG and
+     * `is_unsigned` is `true`, the caller should use the `.as.uint64` field
+     * below to access the properly unsigned value.
+     */
     bool is_unsigned;
 
+    // The MySQL column type of this value.
     TRILOGY_TYPE_t type;
 
+    /* This union is used for accessing the underlying binary type for the value.
+     * Each field member is documented with the MySQL column/value type it maps to.
+     */
     union {
+        /* MySQL types that use this field:
+         *
+         * TRILOGY_TYPE_DOUBLE
+         */
         double dbl;
 
+        /* MySQL types that use this field:
+         *
+         * TRILOGY_TYPE_LONGLONG
+         *
+         * Refer to the `is_unsigned` field above to see which member below should
+         * be used to access the value.
+         */
         int64_t int64;
         uint64_t uint64;
 
+        /* MySQL types that use this field:
+         *
+         * TRILOGY_TYPE_FLOAT
+         */
         float flt;
 
+        /* MySQL types that use this field:
+         *
+         * TRILOGY_TYPE_LONG
+         * TRILOGY_TYPE_INT24
+         *
+         * Refer to the `is_unsigned` field above to see which member below should
+         * be used to access the value.
+         */
         uint32_t uint32;
         int32_t int32;
 
+        /* MySQL types that use this field:
+         *
+         * TRILOGY_TYPE_SHORT
+         *
+         * Refer to the `is_unsigned` field above to see which member below should
+         * be used to access the value.
+         */
         uint16_t uint16;
         int16_t int16;
 
+        /* MySQL types that use this field:
+         *
+         * TRILOGY_TYPE_TINY
+         *
+         * Refer to the `is_unsigned` field above to see which member below should
+         * be used to access the value.
+         */
         uint8_t uint8;
         int8_t int8;
 
+        /* MySQL types that use this field:
+         *
+         * TRILOGY_TYPE_STRING
+         * TRILOGY_TYPE_VARCHAR
+         * TRILOGY_TYPE_VAR_STRING
+         * TRILOGY_TYPE_ENUM
+         * TRILOGY_TYPE_SET
+         * TRILOGY_TYPE_LONG_BLOB
+         * TRILOGY_TYPE_MEDIUM_BLOB
+         * TRILOGY_TYPE_BLOB
+         * TRILOGY_TYPE_TINY_BLOB
+         * TRILOGY_TYPE_GEOMETRY
+         * TRILOGY_TYPE_BIT
+         * TRILOGY_TYPE_DECIMAL
+         * TRILOGY_TYPE_NEWDECIMAL
+         */
         struct {
             const void *data;
             size_t len;
         } str;
 
+        /* MySQL types that use this field:
+         *
+         * TRILOGY_TYPE_DATE
+         * TRILOGY_TYPE_DATETIME
+         * TRILOGY_TYPE_TIMESTAMP
+         */
         struct {
+            /* MySQL types that use this field:
+             *
+             * TRILOGY_TYPE_YEAR
+             */
             uint16_t year;
+
             uint8_t month, day;
         } date;
 
+        /* MySQL types that use this field:
+         *
+         * TRILOGY_TYPE_TIME
+         */
         struct {
             bool is_negative;
             uint32_t days;
--- a/inc/trilogy/protocol.h
+++ b/inc/trilogy/protocol.h
@@ -623,6 +623,7 @@ typedef struct {
 
 typedef struct {
     bool is_null;
+    bool is_unsigned;
 
     TRILOGY_TYPE_t type;
 
--- a/inc/trilogy/protocol.h
+++ b/inc/trilogy/protocol.h
@@ -565,8 +565,6 @@ typedef struct {
 } trilogy_ok_packet_t;
 
 /* trilogy_stmt_ok_packet_t - Represents a MySQL binary protocol prepare response packet.
- *
- * See https://dev.mysql.com/doc/internals/en/com-stmt-prepare-response.html for more detail.
  */
 typedef struct {
     uint32_t id;
@@ -627,7 +625,6 @@ typedef struct {
 
 /* trilogy_binary_value_t - MySQL binary protocol value type
  *
- * See https://dev.mysql.com/doc/internals/en/binary-protocol-value.html for more detail.
  */
 typedef struct {
     // Flag denoting the value is NULL.
--- a/inc/trilogy/protocol.h
+++ b/inc/trilogy/protocol.h
@@ -564,6 +564,10 @@ typedef struct {
     size_t last_gtid_len;
 } trilogy_ok_packet_t;
 
+/* trilogy_stmt_ok_packet_t - Represents a MySQL binary protocol prepare response packet.
+ *
+ * See https://dev.mysql.com/doc/internals/en/com-stmt-prepare-response.html for more detail.
+ */
 typedef struct {
     uint32_t id;
     uint16_t column_count;
--- a/inc/trilogy/protocol.h
+++ b/inc/trilogy/protocol.h
@@ -926,7 +926,7 @@ int trilogy_parse_stmt_ok_packet(const uint8_t *buff, size_t len, trilogy_stmt_o
  * out_values   - Out parameter; A pointer to a pre-allocated array of
  *                trilogy_binary_value_t's. There must be enough space to fit all of the
  *                values. This can be computed with:
- *                `(sizeof(trilogy_value_t) * column_count)`.
+ *                `(sizeof(trilogy_binary_value_t) * column_count)`.
  *
  * Return values:
  *   TRILOGY_OK                   - The packet was was parsed and the out
@@ -934,8 +934,7 @@ int trilogy_parse_stmt_ok_packet(const uint8_t *buff, size_t len, trilogy_stmt_o
  *   TRILOGY_TRUNCATED_PACKET     - There isn't enough data in the buffer
  *                                  to parse the packet.
  *   TRILOGY_PROTOCOL_VIOLATION   - Invalid length parsed for a TIME/DATETIME/TIMESTAMP value.
- *   TRILOGY_UNKNOWN_TYPE         - An unsupported or unknown MySQL type was used in the list
- *                                  of binds.
+ *   TRILOGY_UNKNOWN_TYPE         - An unsupported or unknown MySQL type was seen in the packet.
  *   TRILOGY_EXTRA_DATA_IN_PACKET - There are unparsed bytes left in the
  *                                  buffer.
  */
--- a/inc/trilogy/protocol.h
+++ b/inc/trilogy/protocol.h
@@ -702,7 +702,7 @@ typedef enum {
 int trilogy_build_stmt_execute_packet(trilogy_builder_t *builder, uint32_t stmt_id, uint8_t flags,
                                       trilogy_binary_value_t *binds, uint16_t num_binds);
 
-/* trilogy_build_stmt_bind_data_packet - Build a prepared statement send long data command packet.
+/* trilogy_build_stmt_bind_data_packet - Build a prepared statement bind long data command packet.
  *
  * builder  - A pointer to a pre-initialized trilogy_builder_t.
  * stmt_id  - The statement id for which to build the bind data packet with.
--- a/inc/trilogy/protocol.h
+++ b/inc/trilogy/protocol.h
@@ -565,6 +565,13 @@ typedef struct {
 } trilogy_ok_packet_t;
 
 typedef struct {
+    uint32_t id;
+    uint16_t column_count;
+    uint16_t parameter_count;
+    uint16_t warning_count;
+} trilogy_stmt_ok_packet_t;
+
+typedef struct {
     uint16_t warning_count;
     uint16_t status_flags;
 } trilogy_eof_packet_t;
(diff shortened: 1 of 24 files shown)


Mmkred (approved):

brianmario:
@eileencodes hadn't seen this before, but after merging in your change from #5 it requires approval for the workflow to run on this pull. Make sense I guess. But just FYI for when one of you have a moment 🙏

composerinteralia:
Approved. We've got the "Require approval for first-time contributors" setting selected, so we shouldn't need to do that for you again.

src/protocol.c
```diff
@@ -673,4 +710,459 @@ int trilogy_build_ssl_request_packet(trilogy_builder_t *builder, TRILOGY_CAPABIL
...
+        if (num_binds % 8) {
+            CHECKED(trilogy_builder_write_uint8(builder, current_bits))
+        }
+
+        // new params bound flag
+        CHECKED(trilogy_builder_write_uint8(builder, 0x1));
+
+        for (i = 0; i < num_binds; i++) {
+            CHECKED(trilogy_builder_write_uint8(builder, binds[i].type));
+            CHECKED(trilogy_builder_write_uint8(builder, 0x0));
```
composerinteralia, line 775: I see something about this byte being used to indicate whether the type is unsigned. Is that something we care about here?

composerinteralia (commented):
Looks like I need to approve the workflows every time you push to this branch, so I'll keep doing that. I think once this is merged they will run for you automatically on any future PRs.

I'm working my way through this, trying to make sure I understand everything as I go. Looks great so far!

src/protocol.c
brianmario: Totally missed coming back to this, good catch!

brianmario:
Interesting test failure. Is there a way to re-run individual actions or does it require another push?

composerinteralia:
I've seen that one before. I triggered a re-run

src/client.c
```diff
@@ -725,4 +725,181 @@ void trilogy_free(trilogy_conn_t *conn)
...
+                                             stmt->column_count, values_out);
+    }
+}
+
+int trilogy_stmt_reset_send(trilogy_conn_t *conn, trilogy_stmt_t *stmt)
+{
+    int err = 0;
+
+    trilogy_builder_t builder;
+    err = begin_command_phase(&builder, conn, 0);
```
composerinteralia, line 871: nitpick for consistency, unless I am missing something:

```suggestion
    int err = begin_command_phase(&builder, conn, 0);
```

instead of `int err = 0;` above.

composerinteralia (commented):
I looked through everything and it all seems to check out to me. 🎉 I'll leave it to somebody with more comfort in C than I to give a final review once you are happy with test coverage and such.

Adding this to Ruby gem makes sense to me as a separate PR, since this PR's already got plenty going on.

src/client.c
brianmario: Love it 👍

brianmario:
@composerinteralia amazing thank you for giving it a look over! 🙏 

I'll see if I can find someone to help review the C-specific stuff when I'm ready for it.

composerinteralia:
I played around with some blocking tests (modeled after your client tests) as part of my review. Feel free to use or throw away as necessary!

```c
TEST test_blocking_stmt_prepare()
{
    trilogy_conn_t conn;

    connect_conn(&conn);

    const char *query = "SELECT ?";
    trilogy_stmt_t stmt;

    int err = trilogy_stmt_prepare(&conn, query, strlen(query), &stmt);
    ASSERT_OK(err);

    ASSERT_EQ(1, stmt.parameter_count);

    trilogy_column_packet_t param;
    err = trilogy_read_full_column(&conn, &param);
    ASSERT_OK(err);

    ASSERT_EQ(1, stmt.column_count);

    trilogy_column_packet_t column_def;
    err = trilogy_read_full_column(&conn, &column_def);
    ASSERT_OK(err);

    trilogy_free(&conn);
    PASS();
}

TEST test_blocking_stmt_execute()
{
    trilogy_conn_t conn;

    connect_conn(&conn);

    const char *query = "SELECT ?";
    trilogy_stmt_t stmt;

    int err = trilogy_stmt_prepare(&conn, query, strlen(query), &stmt);
    ASSERT_OK(err);

    ASSERT_EQ(1, stmt.parameter_count);

    trilogy_column_packet_t param;
    err = trilogy_read_full_column(&conn, &param);
    ASSERT_OK(err);

    ASSERT_EQ(1, stmt.column_count);

    trilogy_column_packet_t column_def;
    err = trilogy_read_full_column(&conn, &column_def);
    ASSERT_OK(err);

    uint8_t flags = 0x00;
    const char str[] = {'t', 'e', 's', 't'};
    size_t len = sizeof(str);
    trilogy_binary_value_t binds[] = {
        {.is_null = false, .type = TRILOGY_TYPE_VAR_STRING, .as.str.data = str, .as.str.len = len}};

    uint64_t column_count;

    err = trilogy_stmt_execute(&conn, &stmt, flags, binds, &column_count);
    ASSERT_OK(err);

    ASSERT_EQ(1, column_count);

    trilogy_column_packet_t columns[1];
    err = trilogy_read_full_column(&conn, &columns[0]);
    ASSERT_OK(err);

    trilogy_binary_value_t values[1];
    err = trilogy_stmt_read_full_row(&conn, &stmt, columns, values);
    ASSERT_OK(err);

    ASSERT_MEM_EQ(values[0].as.str.data, "test", values[0].as.str.len);

    err = trilogy_stmt_read_full_row(&conn, &stmt, columns, values);
    ASSERT_EOF(err);

    trilogy_free(&conn);
    PASS();
}

TEST test_blocking_stmt_reset()
{
    trilogy_conn_t conn;

    connect_conn(&conn);

    const char *query = "SELECT ?";
    trilogy_stmt_t stmt;

    int err = trilogy_stmt_prepare(&conn, query, strlen(query), &stmt);
    ASSERT_OK(err);

    ASSERT_EQ(1, stmt.parameter_count);

    trilogy_column_packet_t param;
    err = trilogy_read_full_column(&conn, &param);
    ASSERT_OK(err);

    ASSERT_EQ(1, stmt.column_count);

    trilogy_column_packet_t column_def;
    err = trilogy_read_full_column(&conn, &column_def);
    ASSERT_OK(err);

    err = trilogy_stmt_reset(&conn, &stmt);
    ASSERT_OK(err);

    trilogy_free(&conn);
    PASS();
}

TEST test_blocking_stmt_close()
{
    trilogy_conn_t conn;

    connect_conn(&conn);

    const char *query = "SELECT ?";
    trilogy_stmt_t stmt;

    int err = trilogy_stmt_prepare(&conn, query, strlen(query), &stmt);
    ASSERT_OK(err);

    ASSERT_EQ(1, stmt.parameter_count);

    trilogy_column_packet_t param;
    err = trilogy_read_full_column(&conn, &param);
    ASSERT_OK(err);

    ASSERT_EQ(1, stmt.column_count);

    trilogy_column_packet_t column_def;
    err = trilogy_read_full_column(&conn, &column_def);
    ASSERT_OK(err);

    err = trilogy_stmt_close(&conn, &stmt);
    ASSERT_OK(err);

    trilogy_free(&conn);
    PASS();
}

```

brianmario:
@composerinteralia Thanks! Dropped them in. I'll go back and add some more edge-case tests for everything to try and make sure we've covered some more surface area.

inc/trilogy/protocol.h
```diff
@@ -597,6 +604,128 @@ typedef struct {
     size_t data_len;
 } trilogy_value_t;
 
+typedef struct {
```
dbussink, line 607: Would it be useful to reference https://dev.mysql.com/doc/internals/en/binary-protocol-value.html here to explain where this comes from?

dbussink: Also might be good to annotation each part in the `union` then with which binary protocal value type(s) it implements?

inc/trilogy/protocol.h
```diff
@@ -547,6 +547,13 @@ typedef struct {
     size_t last_gtid_len;
 } trilogy_ok_packet_t;
 
+typedef struct {
```
dbussink, line 550: Adding a reference to https://dev.mysql.com/doc/internals/en/com-stmt-prepare-response.html might be useful here?

inc/trilogy/protocol.h
```diff
@@ -597,6 +604,128 @@ typedef struct {
     size_t data_len;
 } trilogy_value_t;
 
+typedef struct {
+    bool is_null;
+    bool is_unsigned;
```
dbussink, line 609: I think with moving this to the end of the struct and then also moving `is_negative` to the end for date / date time might allow for tighter packing of the struct?

src/builder.c
```diff
@@ -113,6 +113,53 @@ int trilogy_builder_write_uint64(trilogy_builder_t *builder, uint64_t val)
...
+    float f;
+    uint32_t u;
+} trilogy_float_buf_t;
+
+typedef union {
+    double d;
+    uint64_t u;
+} trilogy_double_buf_t;
+
+#endif
```
dbussink, line 129: Should the block above instead be in a header? If it's not, I don't think the define guards would be needed since `*.c` files should only be compiled once?

src/reader.c
```diff
@@ -95,6 +95,53 @@ int trilogy_reader_get_uint64(trilogy_reader_t *reader, uint64_t *out)
...
+    float f;
+    uint32_t u;
+} trilogy_float_buf_t;
+
+typedef union {
+    double d;
+    uint64_t u;
+} trilogy_double_buf_t;
+
+#endif
```
dbussink, line 111: As it's duplicated here, we might want to put it indeed into a header?

inc/trilogy/protocol.h
```diff
@@ -597,6 +604,128 @@ typedef struct {
...
+            const void *data;
+            size_t len;
+        } str;
+
+        struct {
+            uint16_t year;
+            uint8_t month, day;
+            struct {
+                bool is_negative;
+                uint32_t days;
```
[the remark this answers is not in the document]
brianmario: Good call 👍

inc/trilogy/protocol.h
brianmario: Annotating the union is definitely a good idea 👍  It's on my internal list of things to do before taking this out of draft, I'll add it to the pull request body checklist so I don't forget ;)

In general, another documentation pass wouldn't hurt.

src/builder.c
brianmario: To be honest, I kinda hated doing this 😅

But, these types are truly just helpers for the internal serialization implementation(s) and should never have to be used by a public user of this library. I know it's not a big deal to throw it in a header and just mark it private, but as we've both seen with `libmysqlclient` - whatever is exposed to the caller, they'll use and hack in ways we can't imagine.

If this stays in the C files, the guards are because the Ruby extension combines all the C files into a single source file for compilation. That process was actually something I wanted to have as a `Makefile` target for folks to more easily be able to embed this project in their own (which would combine everything into a single `.c`. and `.h`). I haven't checked if it exists yet, but it would be nice if it did.

That said, I could pretty easily be convinced to just put it in a header.

What do you think?

composerinteralia: I think the guards were related to the Ruby library concatenating all the source files: https://github.com/github/trilogy/blob/main/contrib/ruby/ext/trilogy-ruby/extconf.rb#L3-L6

composerinteralia: Ha! You beat me to this comment by a minute

dbussink: @brianmario Would it alternatively work to do this inline? For example:

```c
int trilogy_reader_get_double(trilogy_reader_t *reader, double *out)
    union {double d; uint64_t u;} double_val;
    double_val.d = val;

    int rc = trilogy_reader_get_uint64(reader, &double_val.u);
    if (rc != TRILOGY_OK) {
        return rc;
    }

    *out = double_val.d;
    return TRILOGY_OK;
}
```

That way there's no separate type definition necessary and there's no issue with combining C files either?

brianmario: Ah yeah I like that 👍

test/blocking_test.c
```diff
@@ -155,6 +477,13 @@ int blocking_test()
     RUN_TEST(test_blocking_query);
     RUN_TEST(test_blocking_query_error);
     RUN_TEST(test_blocking_query_no_rows);
+    RUN_TEST(test_blocking_stmt_prepare);
+    RUN_TEST(test_blocking_stmt_execute_str);
+    RUN_TEST(test_blocking_stmt_execute_integer);
+    RUN_TEST(test_blocking_stmt_execute_double);
+    RUN_TEST(test_blocking_stmt_execute_datetime);
```
dbussink, line 484: Would it be possible to create one test for each type that's supported?

brianmario: Yeah that's the plan. I just haven't had a chance to sit down to do it yet 😅

adrianna-chang-shopify:
👋 Hi folks -- I'm gonna pick this up and try to get it across the finish line. @brianmario I can either start a new PR with commits cherry-picked from yours, or if you give me write access to your fork I can push updates here.

Do either you or @composerinteralia remember what sorts of "edge cases" we were still looking to test? (I've yet to take a look at the code in depth, so it's not immediately obvious to me what gaps are left to fill). I'll take a look at the test suite for mysql2's prepared statements for inspiration as well, but figured I'd see if either of y'all could remember where this PR was at 😂

brianmario:
Hi!

@adrianna-chang-shopify I can totally give you access to my fork, not a problem at all.

In terms of where it was left off, it _should_ be ready to go other than my wanting there to be some more tests. I haven't had time to get back to writing them myself, but the rest of the code should be ready to go.

composerinteralia:
I recall this looking fairly ready when I reviewed it a while back. I'd be fine to merging this as-is after a rebase or whatever and then we can fill in whatever additional test coverage we want (e.g. testing with all the different types) separately. Some of that test coverage could probably also be done via the Ruby client once we've added support there.

adrianna-chang-shopify:
Okay, so I've added prepared statement tests for all of the MySQL types, two of which are still failing and have me a bit stumped:
```
Expected: TRILOGY_OK
     Got: TRILOGY_ERR
F
FAIL test_blocking_stmt_execute_year: (TRILOGY_OK) != ((err)) (test/blocking_test.c:758)
```

```
FAIL test_blocking_stmt_execute_float: values[0].as.flt != float_val (test/blocking_test.c:420)
```

The first one I can't manage to repro locally, which is making it difficult to debug. (I apologize for all the noise with the builds, I was trying to debug on CI 😅 ). Seems like building the statement itself is problematic -- if anyone wants to take a stab at reproing locally that would be great 🙏 

The second one _does_ fail locally, but the failure is bizarre. Basically, when Trilogy reads the column back from the resultset, it thinks that the type is `TRILOGY_TYPE_DOUBLE` instead of `TRILOGY_TYPE_FLOAT`. This is shown if print out the value as a float vs a double:
```diff
diff --git a/test/blocking_test.c b/test/blocking_test.c
index 083a132..6da78c0 100644
--- a/test/blocking_test.c
+++ b/test/blocking_test.c
@@ -411,12 +411,16 @@ TEST test_blocking_stmt_execute_float() {
 
     trilogy_column_packet_t columns[1];
     err = trilogy_read_full_column(&conn, &columns[0]);
+    printf("Column type: %d\n", columns[0].type);
     ASSERT_OK(err);
 
     trilogy_binary_value_t values[1];
     err = trilogy_stmt_read_full_row(&conn, &stmt, columns, values);
     ASSERT_OK(err);
 
+    printf("Value as float: %f\n", values[0].as.flt); // wrong
+    printf("Values as double: %lf\n", values[0].as.dbl);
+
     ASSERT_EQ(values[0].as.flt, float_val);
 
     err = trilogy_stmt_read_full_row(&conn, &stmt, columns, values);
```

```
Column type: 5 # TRILOGY_TYPE_DOUBLE
Value as float: 0.000000
Values as double: 1234.500000
F
FAIL test_blocking_stmt_execute_float: values[0].as.flt != float_val (test/blocking_test.c:424)
```

That value is definitely being written correctly:
```diff
diff --git a/src/protocol.c b/src/protocol.c
index 232bb97..21a0577 100644
--- a/src/protocol.c
+++ b/src/protocol.c
@@ -781,6 +781,7 @@ int trilogy_build_stmt_execute_packet(trilogy_builder_t *builder, uint32_t stmt_
         CHECKED(trilogy_builder_write_uint8(builder, 0x1));
 
         for (i = 0; i < num_binds; i++) {
+            printf("Writing type: %d\n", binds[i].type);
             CHECKED(trilogy_builder_write_uint8(builder, binds[i].type));
 
             if (binds[i].is_unsigned) {
```

```
Writing type: 4
Column type: 5 # TRILOGY_TYPE_DOUBLE
Value as float: 0.000000
Values as double: 1234.500000
F
FAIL test_blocking_stmt_execute_float: values[0].as.flt != float_val (test/blocking_test.c:424)
```

But when it's read in `#trilogy_parse_column_packet`, we think it's a double:
```diff
diff --git a/src/protocol.c b/src/protocol.c
index 232bb97..d207ba6 100644
--- a/src/protocol.c
+++ b/src/protocol.c
@@ -379,6 +379,7 @@ int trilogy_parse_column_packet(const uint8_t *buff, size_t len, bool field_list
 
     uint8_t type;
     CHECKED(trilogy_reader_get_uint8(&reader, &type));
+    printf("Read column type: %d\n", type);
     out_packet->type = type;
 
     CHECKED(trilogy_reader_get_uint16(&reader, &out_packet->flags));
```
```
Writing type: 4
Read column type: 5
Column type: 5
Value as float: 0.000000
Values as double: 1234.500000
F
FAIL test_blocking_stmt_execute_float: values[0].as.flt != float_val (test/blocking_test.c:424)
```

Would love input if either of you have it @brianmario @composerinteralia! Maybe I'm missing something simple.

matthewd:
> ```
> Value as float: 0.000000
> Value as double: 1234.500000
> ```

It sounds like it _is_ a double [as opposed to being a float that's mis-tagged]. Perhaps the server accepts incoming floats for compatibility, but just upcasts them to a double immediately?

Ooh, maybe relevant:

> Using FLOAT might give you some unexpected problems because all calculations in MariaDB are done with double precision.

adrianna-chang-shopify:
Oh, good find @matthewd ! I'll dig a bit more -- but yeah, if that's the behaviour on the server side, there's not much we can do there 😄

adrianna-chang-shopify:
Okay, this is finally ready for review 🙈 I'll rebase off main and squash the commits once we deem it merge-able.

brianmario:
Thanks for adding the rest of these tests! 🙏

It looks good to me, but I'll wait for someone with merge access to do a review as well 😉

composerinteralia (approved):

brianmario:
🎉

PR #6, opened 2022-01-04
Fix actions for macos too

eileencodes:
Add pull_request to the list of actions so that PRs opened via forks
will run CI.

Forgot this when I opened #5.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,5 +1,5 @@
 name: macOS
-on: [push]
+on: [push, pull_request]
 
 jobs:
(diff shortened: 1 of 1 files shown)


composerinteralia (approved):

eileencodes:
Merging even though builds aren't passing because they are duplicated. The successful run can be seen here https://github.com/github/trilogy/runs/4703677633?check_suite_focus=true

I think if we want to avoid this in the future project maintainers should fork and only push to their forks. I'm not sure if there's a good solution in actions for this. Will ask around.

PR #8, opened 2022-01-18
Silence clang 13 warnings

casperisfine:
They should probably be addressed, but in the meantime they make the compilation output hard to read.


composerinteralia:
Seems like I can close this PR now. Thank you for fixing these!

PR #9, opened 2022-01-18
Add a bin/console script

casperisfine:
This is really useful when working with the gem.


composerinteralia (approved):
Agreed, thank you!

PR #10, opened 2022-01-18
Allow to adjust read and write timeouts after connecting

casperisfine:
That's a feature of `mysql2` we use often.

### DIFF
--- a/contrib/ruby/test/client_test.rb
+++ b/contrib/ruby/test/client_test.rb
@@ -264,6 +264,55 @@ class ClientTest < TrilogyTest
     ensure_closed client
   end
 
+  def test_adjustable_read_timeout
+    client = new_tcp_client(read_timeout: 5)
+    assert client.query("SELECT SLEEP(0.2)");
+    client.read_timeout = 0.1
+    assert_equal 0.1, client.read_timeout
+    assert_raises Errno::ETIMEDOUT do
+      client.query("SELECT SLEEP(1)")
+    end
+  ensure
+    ensure_closed client
+  end
+
+  def test_read_timeout_closed_connection
+    client = new_tcp_client(read_timeout: 5)
+    client.close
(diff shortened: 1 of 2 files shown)


contrib/ruby/test/client_test.rb
```diff
@@ -264,6 +264,55 @@ def test_read_timeout
...
+  end
+
+  def test_read_timeout_closed_connection
+    client = new_tcp_client(read_timeout: 5)
+    client.close
+    ensure_closed client
+
+    assert_raises IOError do
+      client.read_timeout
+    end
```
casperisfine, line 286: So this is a bit weird, but seemed the most logical.

Alternatively I could return a fallback value, or I could keep a copy of the timeout value in the client object, and return that.

composerinteralia (approved):
Looks good to me. I think erroring once the connection is closed makes sense, unless there is a specific reason to do otherwise.

One rounding quirk that I don't think was caused by anything in this PR but is more obvious now:

Screen Shot 2022-01-18 at 4 17 57 PM

casperisfine:
> I think erroring once the connection is closed makes sense, unless there is a specific reason to do otherwise.

Yeah, I'm not quite sure at this stage. I suppose it's not really expected, so maybe it would make more sense to simply noop?

It's likely that the calling code looks something like:

```ruby
client.read_timeout = something
client.query(...)
```

People might not expect `read_timeout=` to raise (even less so IOError) and only rescue around `query`.

> One rounding quirk

Yeah it's stored as a double, so this is bound to happen, unless we keep the original `VALUE` inside the Ruby object.

composerinteralia:
> Yeah, I'm not quite sure at this stage. I suppose it's not really expected, so maybe it would make more sense to simply noop?

I'll merge as is for now. We can always come back and change it if needed.

PR #15, opened 2022-01-19
Refactor error classes

casperisfine:
Fix: https://github.com/github/trilogy/issues/11

Opening this as a draft because I hit a few roadblocks, so I figured I should solicit some feedback.

The goals are:

  - Not raise any error that isn't a descendant of `Trilogy::Error`.
  - Have a clear distinction between transient network errors that may be retried, and invalid queries that shouldn't.

Unsolved issues (yet)

  - Trilogy calls `rb_syserr_fail_str` in a few places. Meaning it can still raise any of the `Errno::*` errors.
  - `handle_trilogy_error` raise all `TRILOGY_ERR` as `ProtocolError` which is a `ConnectionError`, but
     I'm not certain none of the possible errors could be considered client errors.

cc @composerinteralia @matthewd

### DIFF
--- a/contrib/ruby/lib/trilogy.rb
+++ b/contrib/ruby/lib/trilogy.rb
@@ -1,7 +1,51 @@
-require "trilogy/cext"
 require "trilogy/version"
 
 class Trilogy
+  # Trilogy::Error is the base error type. All errors raised by Trilogy
+  # should be descendants of Trilogy::Error
+  module Error
+  end
+
+  # Trilogy::ConnectionError is the base error type for all potentially transient
+  # network errors.
+  module ConnectionError
+    include Error
+  end
+
+  class BaseError < StandardError
+    include Error
+  end
+
+  # Trilogy::ClientError is the base error type for invalid queries or parameters
+  # that shouldn't be retried.
+  class ClientError < BaseError
+    include Error
+  end
+
+  class QueryError < ClientError
+  end
+
+  class CastError < ClientError
+  end
+
+  class ProtocolError < BaseError
+    include ConnectionError
+
+    attr_reader :error_code, :error_message
+  end
+
+  class SSLError < BaseError
+    include ConnectionError
+  end
+
+  class ConnectionClosed < IOError
+    include ConnectionError
+  end
+
+  class TimeoutError < Errno::ETIMEDOUT
+    include ConnectionError
+  end
+
   def in_transaction?
     (server_status & SERVER_STATUS_IN_TRANS) != 0
   end
@@ -28,34 +72,38 @@ class Trilogy
   ensure
     self.query_flags = old_flags
   end
-end
 
-Trilogy::Result.class_eval do
-  def count
-    rows.count
-  end
+  class Result
+    attr_reader :fields, :rows, :query_time
 
-  def each_hash
-    return enum_for(:each_hash) unless block_given?
-
-    rows.each do |row|
-      this_row = {}
-
-      idx = 0
-      row.each do |col|
-        this_row[fields[idx]] = col
-        idx += 1
-      end
-
-      yield this_row
+    def count
+      rows.count
     end
 
-    self
-  end
+    def each_hash
+      return enum_for(:each_hash) unless block_given?
 
-  def each(&bk)
-    rows.each(&bk)
-  end
+      rows.each do |row|
+        this_row = {}
 
-  include Enumerable
+        idx = 0
+        row.each do |col|
+          this_row[fields[idx]] = col
+          idx += 1
+        end
+
+        yield this_row
+      end
+
+      self
+    end
+
+    def each(&bk)
+      rows.each(&bk)
+    end
+
+    include Enumerable
+  end
 end
+
+require "trilogy/cext"
--- a/contrib/ruby/lib/trilogy.rb
+++ b/contrib/ruby/lib/trilogy.rb
@@ -28,10 +28,42 @@ class Trilogy
   class CastError < ClientError
   end
 
-  class ProtocolError < BaseError
+
+  class TimeoutError < Errno::ETIMEDOUT
     include ConnectionError
+  end
+
+  class ProtocolError < BaseError
+    ERROR_CODES = {
+      1205 => TimeoutError, # ER_LOCK_WAIT_TIMEOUT
+      1044 => ConnectionError, # ER_DBACCESS_DENIED_ERROR
+      1045 => ConnectionError, # ER_ACCESS_DENIED_ERROR
+      1152 => ConnectionError, # ER_ABORTING_CONNECTION
+      1153 => ConnectionError, # ER_NET_PACKET_TOO_LARGE
+      1154 => ConnectionError, # ER_NET_READ_ERROR_FROM_PIPE
+      1155 => ConnectionError, # ER_NET_FCNTL_ERROR
+      1156 => ConnectionError, # ER_NET_PACKETS_OUT_OF_ORDER
+      1157 => ConnectionError, # ER_NET_UNCOMPRESS_ERROR
+      1158 => ConnectionError, # ER_NET_READ_ERROR
+      1159 => ConnectionError, # ER_NET_READ_INTERRUPTED
+      1160 => ConnectionError, # ER_NET_ERROR_ON_WRITE
+      1161 => ConnectionError, # ER_NET_WRITE_INTERRUPTED
+      1927 => ConnectionError, # ER_CONNECTION_KILLED
+    }
 
     attr_reader :error_code, :error_message
+
+    class << self
+      def from_code(message, code)
+        ERROR_CODES.fetch(code, self).new(message, code)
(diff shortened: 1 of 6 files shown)


adrianna-chang-shopify:
@jhawthorn @composerinteralia @matthewd we're looking to get the ball rolling on this work again. Do y'all have any feedback in the meantime? Thanks ❤️ 

cc @paarthmadan

composerinteralia:
I can take a look early next week if nobody beats me to it. I vaguely remember liking this change, if that helps at all 😄.

adrianna-chang-shopify:
Hey @composerinteralia , thank you for the feedback! 🙇‍♀️ 

> My only concern is compatibility with

Given @byroot 's approach using modules, I think backwards compatibility should be retained, since all of the error classes will still inherit from the original error classes. In moving towards _just_ `Trilogy:Error` classes, would adding more extensive test coverage for `LostConnectionExceptionTranslator` help mitigate the risk? 

> one custom bit of translation we've got at GitHub for retrying timeouts on connect

I'm curious why this translation bit wasn't upstreamed into the AR adapter for Trilogy. I thought it was odd that underlying errors like `Errno::ECONNREFUSED` were being translated to `ActiveRecord::StatementInvalid`. Does it make sense to get this upstreamed first before we start changing the error classes?

>> Trilogy calls rb_syserr_fail_str in a few places. Meaning it can still raise any of the Errno::* errors.

> I know we've got adapter code that rescues and translates EPIPE, ETIMEDOUT, ECONNREFUSED, and ECONNRESET, so having Trilogy-specific for those in particular seems helpful.

@byroot do you remember -- was the issue here that we're calling `rb_syserr_fail_str` in places where we're not sure what the appropriate wrapper error would be? 🤔 Were you looking for feedback on what the `Trilogy::Error` mapping should be?

> The error_code on that is 1064, a query parse error. There would be no reason to retry that.
In theory we could raise different errors for different error_codes, but I'm not sure how practical that is since there are lots of them.

Would it make sense to do something more similar to what [mysql2 does](https://github.com/brianmario/mysql2/blob/master/lib/mysql2/error.rb#L15-L27) and map explicit error codes to `ProtocolError` / `ConnectionError`, and have everything else raised as a more generic `BaseError`?

casperisfine:
> was the issue here that we're calling `rb_syserr_fail_str` in places where we're not sure what the appropriate wrapper error would be?

It's been a while. But the issue with `rb_syserr_fail_str` is that you pass an `errno`, so you need to carefully check the manpage of each of the function that `errno` might come from to see which are the possible raised exception. And that can sometime differ from platform to platform. Overall it could raise any of `SystemCallError.subclasses` which contains `107` different errors.

careful auditing might reduce this to a handful, but it's a lot of work and error prone.

composerinteralia:
> Given @byroot 's approach using modules, I think backwards compatibility should be retained, since all of the error classes will still inherit from the original error classes.

👍🏻 I'm probably unnecessarily worried.

> I'm curious why this translation bit wasn't upstreamed into the AR adapter for Trilogy.

Good question. I think we were in a rush to fix a bug and then never revisited it. From my commit message for that change:

> We could possibly solve this in Rails instead, and we should probably
change Trilogy to raise some more easily rescuable errors so we don't
need this, but making this change in <internal library> seems like the
quickest way to get something deployable.

> Would it make sense to ... map explicit error codes to ProtocolError / ConnectionError, and have everything else raised as a more generic BaseError?

Yeah, I think that'd be OK, especially if we have a well-known list we can draw from.

adrianna-chang-shopify:
Okay, so following the feedback from our last conversation, we've:
- Brought in the error code mappings from mysql2, so that certain `TRILOGY_ERR`s are raised as `ConnectionError` or `TimeoutError`, with remaining `TRILOGY_ERR`s being considered `ProtocolError`s.
- Decided to hold off on touching places where we call `rb_syserr_fail_str` for now. We can do this in a second pass.

contrib/ruby/lib/trilogy.rb
```diff
@@ -1,7 +1,79 @@
...
+
+
+  class TimeoutError < Errno::ETIMEDOUT
+    include ConnectionError
+  end
+
+  class ProtocolError < BaseError
+    ERROR_CODES = {
+      1205 => TimeoutError, # ER_LOCK_WAIT_TIMEOUT
+      1044 => ConnectionError, # ER_DBACCESS_DENIED_ERROR
```
casperisfine, line 39: I just realized. We can't raise `ConnectionError` as it's a module. We'll need to create another error that include that module.

We should also probably try to cover that path.

composerinteralia:
> it would be good to have a conversation about how much leeway we have to make breaking changes to translating errors in the adapter / what that process should look like.

I think it's fine to make breaking changes—there's probably not too much use of this library outside GitHub yet. It'd be helpful to get a corresponding activerecord-trilogy-adapter branch that passes with these changes so we can run it all against our CI before merging.

adrianna-chang-shopify:
>  It'd be helpful to get a corresponding activerecord-trilogy-adapter branch that passes with these changes 

I can absolutely do that!

composerinteralia:
@haldun is running these changes against our CI. We'll let you know if anything comes up.

composerinteralia:
GitHub CI passing after renaming a couple things 👍🏻

contrib/ruby/lib/trilogy.rb
```diff
@@ -28,34 +109,38 @@ def query_with_flags(sql, flags)
...
     self.query_flags = old_flags
   end
-end
 
-Trilogy::Result.class_eval do
-  def count
-    rows.count
-  end
+  class Result
+    attr_reader :fields, :rows, :query_time
```
composerinteralia, line 114: Just noticed these warnings in the activerecord-trilogy-adapter tests:

```
warning: method redefined; discarding old fields
warning: method redefined; discarding old rows
warning: method redefined; discarding old query_time
```

Defining these in Ruby seems good to me, but then I assume we'll need to remove the `rb_define_attr`s.

adrianna-chang-shopify: I'll look into this! 👀

adrianna-chang-shopify: Done!

composerinteralia:
CI keeps getting cancelled (timing out?) for some reason. Once it's green I'll merge.

casperisfine:
Thanks. Applied in [1a9ab90](https://github.com/github/trilogy/pull/15/commits/1a9ab9052999e33248dce0edeb8f51212064f1bd)

composerinteralia:
Looks like `test_connection_error` is flaky—sometimes we end up with a TRILOGY_PROTOCOL_VIOLATION. I tried running the same tests on main and saw the same flakiness, so I think the new test is exposing an existing bug.

I'm going to merge and look into that separately. If the flakiness is disruptive in the meantime we may need to remove that test.

composerinteralia (approved):

lorint:
> I think the new test is exposing an existing bug

(Hi - I'm Lorin Thwaits, the guy adding [AR 6.1 and 7.0 compatibility to Trilogy](https://github.com/github/activerecord-trilogy-adapter/pull/20).  Long time litener / first time caller!)

The Python folks had this same PROTOCOL_VIOLATION error unexpectedly surface starting with the release of MySQL 8.0.24 -- a [MySQL bug had been fixed](https://bugs.mysql.com/bug.php?id=101479) at that point and the change notes for that release indicate:
"Prepared statements without parameters violate the MySQL protocol by sending unnecessary extra bytes."

adrianna-chang-shopify:
👋 Hi @lorint , thanks for supplying some extra context! Do you know if the Python folks were experiencing flakiness as well, or was the `PROTOCOL_VIOLATION` error showing up consistently after moving to MySQL 8.0.24? 

I suspect this might be something different though, I don't believe Trilogy supports prepared statements yet 🤔

lorint:
> Do you know if the Python folks were experiencing flakiness as well?

I think it was a consistent failure they were seeing.

With this new patch in place then running our suite on 8.0 I get a perfectly consistent failure.  Somewhat fast and multithreaded machine -- Macbook M1.

adrianna-chang-shopify:
> With this new patch in place then running our suite on 8.0 I get a perfectly consistent failure.

Do you have an example of a test that's failing consistently from your suite? I can't get `test_connection_error` to fail locally on my end, even with MySQL 8 + Ruby 3.2. I'm also on an M1.

lorint:
Now that I went back to check this out I see that I have mis-spoken!  The failure I had seen was after I had merged in your ARTA PR "Translate new error classes" along with my compatibility PR, on this line:
https://github.com/adrianna-chang-shopify/activerecord-trilogy-adapter/blob/ac-translate-new-error-classes/test/lib/active_record/connection_adapters/trilogy_adapter_test.rb#L44

But now it's been fixed and merged upstream here in this PR on the Trilogy gem.  Have just done the same thing again and everything works :)

PR #17, opened 2022-01-28
Include more post-query info on Result

matthewd:
Instead of leaving the caller to read `last_insert_id` and `affected_rows` from the connection before they're overwritten, let's copy them to the returned Result object.

While the connection always returns integers for those values (`0` when they're N/A), it seems more reasonable to assume callers anticipate the shape of their particular result, so here I'm instead preferring to return `nil` on results for queries where these don't apply. (As a side-effect this allows a caller to distinguish a zero-row-returning SELECT (`rows == []`, `affected_rows == nil`) from a zero-row-affecting UPDATE (`rows == []`, `affected_rows == 0`). I'm not sure where that might be specifically useful, but it seems nice to have?)


composerinteralia (approved):
I like it. Seems convenient to have these available on the result, and the `nil` cases make sense to me.

Issue #26, opened 2022-09-12
Having caching_sha2_password authentication issues with mysql 8.0.30


abdulchaudhrycoupa:
Everything works, in that I don't see any issues with trilogy when running with `mysql_native_password` auth but the moment I switch to `caching_sha2_password` things break down in weird ways.

I am trying to understand where it fails but I am still learning how this works and I am only seeing a packet error.

It seems to perform authentication and create a successful connection to the mysql server when switching the client over to `caching_sha2_password` but then it fails during packet parsing for some reason.

I can see in `show processlist` during the handshake that the user moves from unauthorized to the user name but then it fails on the client just as its about to call `trilogy_auth_recv` or maybe its calling it too often with the wrong packet sequence/payload and the server disconnects. I am still debugging the issue.

I am using a debug version of the mysql server version `8.0.30-debug` and I was about to debug the server side to see if it was indeed closing the connection because it didn't like something from the client.

Is this a known issue with `caching_sha2_password` and trilogy?

abdulchaudhrycoupa:
@composerinteralia just pulled the latest, that doesn't seem to help. The error that I am seeing is `TRILOGY_UNEXPECTED_PACKET`

composerinteralia:
It seems like I'm able to reproduce this (or at least something like it). I pushed up [a failing test](https://github.com/github/trilogy/commit/824af709468143e63bef08c1526244f421cd8f2c) and do see a [TRILOGY_UNEXPECTED_PACKET error](https://github.com/github/trilogy/actions/runs/3184907843/jobs/5193858582). Does this capture the problem you are having?

abdulchaudhrycoupa:
@composerinteralia that does reproduce the problem exactly. 

created PR - https://github.com/github/trilogy/pull/28

This seems to fix the test but do look into it more detail as you might want to capture the payload that the server is sending, in this case it just seems to be the plugin information which we don't care about that much, for other schemes, such as FIDO it might be more detailed.

composerinteralia:
Thanks for digging into that! The tests aren't green, but the error has changed to `trilogy_auth_recv: 1159 Got timeout reading communication packets`, so that's something.

abdulchaudhrycoupa:
@composerinteralia it seems to fix the test suite for me for the mysql server version that I am using, but I have a very strange version of the mysql server, its `8.0.30-debug` as it built from source in debug mode, it might be breaking on other versions.

breisig:
[...]

PR #28, opened 2022-10-04
caching_sha returns auth_more_data

abdulchaudhrycoupa:
The caching_sha2_password authentication sends auth_more_data, the data doesn't seem to be that useful, its just the current authentication plugin but for other authentication schemes it might include more useful information.

This seems to fix issue #26 

@composerinteralia - I believe this fixes your broken test. From the documentation it seems only mysql_native_password returns ok, the other non challenge-response auth plugin schemes return auth more data.

See https://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::AuthMoreData

Ideally, this should capture the data payload but for now seems to pass tests.

### DIFF
--- a/inc/trilogy/protocol.h
+++ b/inc/trilogy/protocol.h
@@ -367,6 +367,7 @@ typedef enum {
 // Typical response packet types
 typedef enum {
     TRILOGY_PACKET_OK = 0x0,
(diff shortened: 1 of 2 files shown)


abdulchaudhrycoupa:
Closing this PR

PR #30, opened 2022-10-05
Add #closed? method to Trilogy client

paarthmadan:
This PR extends the API available on the Ruby bindings to include a `#closed?` predicate method. It's use-case is inspired by the ongoing work to introduce a [Semian adapter for Trilogy](https://github.com/Shopify/semian/pull/420).

It uses the presence of the `conn.socket` to make the conclusion of whether the connection is currently closed or opened—this is the same condition that `rb_trilogy_close` uses to determine whether or not to initiate a close request, so I'd assume it's sufficient. Open to hearing feedback on this, though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+  - Add `#closed?` method to Ruby binding. #30
+
 ## 2.1.2
 
 2022-10-04
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -826,6 +826,17 @@ static VALUE rb_trilogy_close(VALUE self)
     return Qnil;
 }
 
+static VALUE rb_trilogy_closed(VALUE self)
+{
+    struct trilogy_ctx *ctx = get_ctx(self);
+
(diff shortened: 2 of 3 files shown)


jhawthorn:
In what cases do you expect `closed?` to return true? I think with this implementation it's likely only happening when `closed` is called explicitly on the connection, but not if it has been closed due to an error (including in cases where `TRILOGY_CLOSED_CONNECTION` error is returned). Is that still helpful to your use case?

casperisfine:
> ? I think with this implementation it's likely only happening when `closed` is called explicitly on the connection

Yes, that's expected. The same than `Mysql2::Client#closed?` https://github.com/brianmario/mysql2/blob/ba4d46551d132492b34205cdb8fa224c92765bef/ext/mysql2/client.c#L522-L525

> Is that still helpful to your use case?

Yes.

> I looked through the Semian PR, but didn't see any discussion of why closed? was necessary

Yeah, the reference in in a code comment, you can see https://github.com/Shopify/semian/pull/396 for context and also https://github.com/rails/rails/pull/45981

But in short Semian need to know wether an error is caused by a network / server issue. Since many methods raise when called on an explictly closed connection, we don't want to open the circuit breakers in such case. So being able to check `closed?` is helpful for us.

```ruby
>> Trilogy.new({}).tap(&:close).ping
(irb):3:in `ping': connection closed (IOError)
```

Otherwise IOError is too generic, and we'd have to check the error message which I don't like to depend on.

jhawthorn:
👍 Thanks for the explanation. Works for me

PR #31, opened 2022-10-10
ruby: Use TypedData_* APIs

casperisfine:
The `Data_*` API still works, and AFAIK are not officially deprecated but they're certainly no longer recommended.

In the context of trilogy the difference in implementation is very marginal.

As it doesn't hold onto any `VALUE` reference, it can register itself as write barrier protected, so it isn't marked on every minor GC runs.

cc @jhawthorn

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -31,10 +31,38 @@ struct trilogy_ctx {
     unsigned int query_flags;
 };
 
+static void free_trilogy(void *ptr)
+{
+    struct trilogy_ctx *ctx = ptr;
+    if (ctx->conn.socket != NULL) {
+        trilogy_free(&ctx->conn);
+    }
+    xfree(ptr);
+}
+
+static size_t trilogy_memsize(const void *ptr) {
+    const struct trilogy_ctx *ctx = ptr;
+    size_t memsize = sizeof(struct trilogy_ctx);
+    if (ctx->conn.socket != NULL) {
+        memsize += sizeof(trilogy_sock_t);
+    }
+    return memsize;
+}
+
+const rb_data_type_t trilogy_data_type = {
+    .wrap_struct_name = "trilogy",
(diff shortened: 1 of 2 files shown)


contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -91,20 +113,11 @@ static void handle_trilogy_error(struct trilogy_ctx *ctx, int rc, const char *ms
...
-    }
-}
-
 static VALUE allocate_trilogy(VALUE klass)
 {
     struct trilogy_ctx *ctx;
 
-    VALUE obj = Data_Make_Struct(klass, struct trilogy_ctx, NULL, free_trilogy, ctx);
-
-    memset(ctx->server_version, 0, sizeof(ctx->server_version));
```
casperisfine, line 107: I don't think this `memset` was ever necessary since `Make_Struct` use `ZALLOC`.

contrib/ruby/test/client_test.rb
```diff
@@ -638,4 +638,10 @@ def test_connection_invalid_dns
     end
     assert_equal "trilogy_connect - unable to connect to mysql.invalid:3306: TRILOGY_DNS_ERR", ex.message
   end
+
+  def test_memsize
+    require 'objspace'
+    client = new_tcp_client
+    assert_instance_of Integer, ObjectSpace.memsize_of(client)
```
casperisfine, line 645: I was a bit surprised here because:

```ruby
>> ObjectSpace.memsize_of(client)
=> 33312
```

Which seems huge.

casperisfine: Nevermind, most of that is explained by:

```c
#define TRILOGY_DEFAULT_BUF_SIZE 32768
```

jhawthorn:
rebased to include #33 and use `kind_of Integer` for ancient Ruby support (though we should probably drop the stretch build now that it is EOL)

jhawthorn (approved):

PR #33, opened 2022-10-14
Fix memory leak

composerinteralia:
@jhawthorn noticed that we weren't freeing the trilogy_ctx. I tested locally with:

```rb
#!/usr/bin/env ruby

require "bundler/setup"
require "trilogy"

1000000.times { GC.start; Trilogy.new(database: "test").close }
```

and can confirm that RSS climbs consistently before this patch, and does not climb once this patch is applied.

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -96,6 +96,7 @@ static void free_trilogy(struct trilogy_ctx *ctx)
     if (ctx->conn.socket != NULL) {
         trilogy_free(&ctx->conn);
     }
+    xfree(ctx);
 }
 
 static VALUE allocate_trilogy(VALUE klass)


contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -96,6 +96,7 @@ static void free_trilogy(struct trilogy_ctx *ctx)
     if (ctx->conn.socket != NULL) {
         trilogy_free(&ctx->conn);
     }
+    free(ctx);
```
jhawthorn, line 99: I think this should be Ruby's `xfree`

composerinteralia: 👍🏻 Thanks! I can't say I fully understand why yet, but I see that it's doing a whole bunch of extra stuff.

jhawthorn (approved):

brasic:
I wonder if there is value in running valgrind in CI on trilogy (and other c extensions in use in dotcom)

PR #37, opened 2022-11-23
Cast to Integer when there are no decimals

composerinteralia:
Prior to this commit we'd always cast a SUM query to `BigDecimal`, since the column type we get from the server is 246 (i.e. TRILOGY_TYPE_NEWDECIMAL).

After this commit we'll only cast to `BigDecimal` if the column includes decimal digits. If there are none (i.e. if column->decimals is 0), we cast to `Integer` instead.

Closes https://github.com/github/trilogy/issues/36

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cast.c
+++ b/contrib/ruby/ext/trilogy-ruby/cast.c
@@ -142,10 +142,14 @@ rb_trilogy_cast_value(const trilogy_value_t *value, const struct column_info *co
         }
         case TRILOGY_TYPE_DECIMAL:
         case TRILOGY_TYPE_NEWDECIMAL: {
-            // TODO - optimize so we don't have to allocate a ruby string for
-            // decimal columns
-            VALUE str = rb_str_new(value->data, value->data_len);
-            return rb_funcall(rb_mKernel, id_BigDecimal, 1, str);
+            if (column->decimals == 0) {
+                return rb_cstr2inum(value->data, 10);
+            } else {
+                // TODO - optimize so we don't have to allocate a ruby string for
+                // decimal columns
+                VALUE str = rb_str_new(value->data, value->data_len);
+                return rb_funcall(rb_mKernel, id_BigDecimal, 1, str);
+            }
         }
         case TRILOGY_TYPE_FLOAT:
         case TRILOGY_TYPE_DOUBLE: {
--- a/contrib/ruby/ext/trilogy-ruby/cast.c
+++ b/contrib/ruby/ext/trilogy-ruby/cast.c
@@ -7,7 +7,7 @@
 
 #define CAST_STACK_SIZE 64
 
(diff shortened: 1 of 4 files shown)


contrib/ruby/ext/trilogy-ruby/cast.c
```diff
@@ -144,8 +144,12 @@ rb_trilogy_cast_value(const trilogy_value_t *value, const struct column_info *co
         case TRILOGY_TYPE_NEWDECIMAL: {
             // TODO - optimize so we don't have to allocate a ruby string for
             // decimal columns
-            VALUE str = rb_str_new(value->data, value->data_len);
-            return rb_funcall(rb_mKernel, id_BigDecimal, 1, str);
+            if (column->decimals == 0) {
+              return rb_cstr2inum(value->data, 10);
+            } else {
+              VALUE str = rb_str_new(value->data, value->data_len);
```
matthewd, line 150: This file uses 4 space indent.

Also tiny, but as you're touching it anyway, the above TODO can come down into this else clause -- in the integer case, with this change it's to-done.

matthewd:
Just for my own future re-reading, I'll note that this is branching on whether the returned column type has any decimal _digits_ -- i.e. `DECIMAL(10,2)` vs `DECIMAL(10,0)` -- rather than anything to do with the specific row _values_ being returned.

composerinteralia:
> Just for my own future re-reading, I'll note that this is branching on whether the returned column type has any decimal digits -- i.e. DECIMAL(10,2) vs DECIMAL(10,0) -- rather than anything to do with the specific row values being returned.

Thank you for reading what I meant instead of what I wrote 🤦🏻. I'll update it to say digits.

contrib/ruby/ext/trilogy-ruby/cast.c
```diff
@@ -142,10 +142,14 @@ rb_trilogy_cast_value(const trilogy_value_t *value, const struct column_info *co
         }
         case TRILOGY_TYPE_DECIMAL:
         case TRILOGY_TYPE_NEWDECIMAL: {
-            // TODO - optimize so we don't have to allocate a ruby string for
-            // decimal columns
-            VALUE str = rb_str_new(value->data, value->data_len);
-            return rb_funcall(rb_mKernel, id_BigDecimal, 1, str);
+            if (column->decimals == 0) {
+                return rb_cstr2inum(value->data, 10);
```
composerinteralia, line 146: @hparker asked me some questions about value->data and now I wonder whether this should be `rb_cstr_parse_inum`, which takes a length.

composerinteralia: Well not that, since I don't think it's public. But maybe the same thing we do in the else branch?

HParker (approved):
Good choice saving the optimization for later!

PR #40, opened 2023-01-11
Implement `#abandon_results!` for Ruby binding

paarthmadan:
#### What?

This PR adds a method for the Ruby binding: `#abandon_results!`

#### Why?

Following the introduction of multi-statement support (#35), a common pattern for clients is to discard the queue of result sets so as to put the client back into a usable state as quickly as possible. This is especially helpful when multi-statement is being used for bulk inserts, where the result sets aren't particularly relevant. This helper method simplifies that pattern.

#### Details

I opted to return a count signifying the number of results abandoned. I think this can be helpful and doubles as a non-zero success status.

#### Better compatibility with `mysql2`

This method exists on `mysql2` and serves the same purpose.

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -764,6 +764,20 @@ static VALUE rb_trilogy_more_results_exist(VALUE self)
     }
 }
 
+static VALUE rb_trilogy_abandon_results(VALUE self)
+{
+    struct trilogy_ctx *ctx = get_open_ctx(self);
+
+    int count = 0;
+
+    while(ctx->conn.server_status & TRILOGY_SERVER_STATUS_MORE_RESULTS_EXISTS) {
+        VALUE result = execute_read_query_response(ctx);
+        count++;
+    }
+
+    return INT2NUM(count);
+}
+
 static VALUE rb_trilogy_query(VALUE self, VALUE query)
 {
     struct trilogy_ctx *ctx = get_open_ctx(self);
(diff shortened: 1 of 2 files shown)


contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -764,6 +764,20 @@ static VALUE rb_trilogy_more_results_exist(VALUE self)
...
 }
 
+static VALUE rb_trilogy_abandon_results(VALUE self)
+{
+    struct trilogy_ctx *ctx = get_open_ctx(self);
+
+    int count = 0;
+
+    while(ctx->conn.server_status & TRILOGY_SERVER_STATUS_MORE_RESULTS_EXISTS) {
+        VALUE result = execute_read_query_response(ctx);
```
paarthmadan, line 774: I benchmarked this method with an implementation written in pure Ruby using `more_results_exist?` and `next_result` to achieve the same thing. There was no significant difference (which is mostly expected).

I'm curious if there's any optimizations we should make _knowing_ we never actually needed the `result` to be passed back. Should I free the `result` here?

brianmario:
If I recall correctly, one of the main reasons we decided not to add a method like this was because it ended up being "cheaper" to just throw away (close) the connection and open a new one. Doing so lets the server to just free the result set memory server-side and skip sending anything over the wire. This may not be the case anymore, but I have a memory of that.

If the current maintainers are interested in this addition: I think it could be made a little lower level using only the trilogy C API to skip creating any Ruby objects at all (and thus any potential GC pressure from doing that). The [read_query_response](https://github.com/github/trilogy/blob/main/contrib/ruby/ext/trilogy-ruby/cext.c#L583) code would be a good starting place for how to do that.

paarthmadan:
Closing this pull request for now:

> "cheaper" to just throw away (close) the connection and open a new one

This seems to be a reasonable assertion. We can circle back on introducing this into the API if and when it becomes a more tangible concern.

PR #41, opened 2023-01-13
Error class refactoring (part 2)

adrianna-chang-shopify:
Follow up to: https://github.com/github/trilogy/pull/15

Converts a few more Trilogy errors:
- `TRILOGY_DNS_ERROR` is generally returned when we attempt to connect to a db with an unknown hostname, which we should treat as a connection error and handle properly [in the adapter](https://github.com/github/activerecord-trilogy-adapter/blob/main/lib/active_record/connection_adapters/trilogy_adapter.rb#L98-L99).
- `Trilogy::ConnectionClosed` was introduced in the last PR to handle `IOError`, but I forgot to actually raise the Trilogy error in the C ext, so that's been updated
- Handles known syscall errors: `ECONNREFUSED` and `ECONNRESET`, which should both be handled as connection errors. These are the ones I've been encountering frequently; we can expand on the error conversions as we see more pop up. I'm not sure how many of the 107 syscall errors we actually expect to see.

I plan to update https://github.com/github/activerecord-trilogy-adapter/pull/24 to handle these changes.

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -15,7 +15,8 @@
 #define TRILOGY_RB_TIMEOUT 1
 
 VALUE Trilogy_CastError;
-static VALUE Trilogy_ProtocolError, Trilogy_Result, Trilogy_SSLError, Trilogy_QueryError, Trilogy_TimeoutError;
+static VALUE Trilogy_BaseConnectionError, Trilogy_ProtocolError, Trilogy_SSLError, Trilogy_QueryError,
+    Trilogy_TimeoutError, Trilogy_Result;
 
 static ID id_socket, id_host, id_port, id_username, id_password, id_found_rows, id_connect_timeout, id_read_timeout,
     id_write_timeout, id_keepalive_enabled, id_keepalive_idle, id_keepalive_interval, id_keepalive_count,
@@ -108,6 +109,10 @@ static void handle_trilogy_error(struct trilogy_ctx *ctx, int rc, const char *ms
             trilogy_sock_shutdown(ctx->conn.socket);
         }
         rb_raise(Trilogy_SSLError, "%" PRIsVALUE ": SSL Error: %s", rbmsg, ERR_reason_error_string(ossl_error));
+    }
+
+    case TRILOGY_DNS_ERR: {
+        rb_raise(Trilogy_BaseConnectionError, "%" PRIsVALUE ": TRILOGY_DNS_ERROR", rbmsg);
     }
 
     default:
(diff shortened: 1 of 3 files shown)


contrib/ruby/test/client_test.rb
```diff
@@ -369,7 +369,7 @@ def test_trilogy_closed?
   def test_read_timeout
     client = new_tcp_client(read_timeout: 0.1)
 
-    assert_raises Errno::ETIMEDOUT do
+    err = assert_raises Trilogy::TimeoutError do
```
paarthmadan, line 372: Nit: Doesn't look like `err` is used just yet, perhaps we assert on the message or remove the variable assignment

paarthmadan (commented):
🎉

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -995,6 +1014,12 @@ void Init_cext()
     Trilogy_TimeoutError = rb_const_get(Trilogy, rb_intern("TimeoutError"));
     rb_global_variable(&Trilogy_TimeoutError);
 
+    Trilogy_BaseConnectionError = rb_const_get(Trilogy, rb_intern("BaseConnectionError"));
```
jhawthorn, line 1017: I should have given this feedback in the previous PR 😓, but should we rename this `ConnectionError`? I find the "Base" prefix a little out of place as nothing inherits this.

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -70,7 +71,7 @@ static struct trilogy_ctx *get_open_ctx(VALUE obj)
     struct trilogy_ctx *ctx = get_ctx(obj);
 
     if (ctx->conn.socket == NULL) {
-        rb_raise(rb_eIOError, "connection closed");
+        rb_raise(Trilogy_ConnectionClosedError, "Trilogy connection closed");
```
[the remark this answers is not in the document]
adrianna-chang-shopify: Yep SGTM! 👍

adrianna-chang-shopify:
Still seeing pretty consistent failures for `test_connection_error` with MySQL 8.0 🤔 I haven't been able to repro locally though. Any progress on that end @composerinteralia ?

composerinteralia:
Or maybe this? https://github.com/github/trilogy/pull/42

adrianna-chang-shopify:
@composerinteralia @jhawthorn would love to get your help to move this along!
- For now, I've skipped the `test_connection_error` (with a note linking to https://github.com/github/trilogy/pull/42, which is where most of the context is tracked so far) until we have a chance to investigate this further. I'm wondering if it could possibly be OS related.. I'm on `arm64-darwin21` and _still_ can't replicate the failures on my end, whereas CI is `x86_64-darwin21` (not sure what your machine is, Daniel).
- @jhawthorn I know you had some concerns around the naming in https://github.com/github/trilogy/pull/41#discussion_r1070092177, what are your thoughts on the current "module-approach" to preserve backwards compat?
- One thing I noticed and could use feedback on: `TRILOGY_CLOSED_CONNECTION` is currently surfaced as a `QueryError`, and that seems potentially incorrect to me. Do we think this should be a `Trilogy::ConnectionClosed` error?

eileencodes (approved):

HParker (approved):

PR #42, opened 2023-01-17
Avoid protocol error on empty password auth switch

composerinteralia:
Prior to this commit `test_connection_error`, which creates a connection with a username that doesn't exist and an empty password, was flaky on mysql 8. Sometimes it passed with the expected BaseConnectionError, but often it failed with a `TRILOGY_PROTOCOL_VIOLATION`.

The source of the `TRILOGY_PROTOCOL_VIOLATION` was `read_auth_switch_packet`, which only handles `mysql_native_password` and `caching_sha2_password`, whereas in the case of this failure the attempted auth switch was to `sha256_password`.

I'm not entirely sure why we get an auth switch packet sometimes and an "access denied" error packet other times (I'm guessing it has something to do with mysql 8 changing the default auth plugin to the caching plugin, but that's really just a guess). But in the case of an empty password it seems like we can handle the switch just fine. We've done a similar password_length check in https://github.com/github/trilogy/pull/21.

With this commit test_connection_error passes consistently.

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -259,10 +259,12 @@ static int read_auth_switch_packet(trilogy_conn_t *conn, trilogy_handshake_t *ha
         return rc;
     }
 
-    if (strcmp("mysql_native_password", auth_switch_packet.auth_plugin) &&
-        strcmp("caching_sha2_password", auth_switch_packet.auth_plugin)) {
-        // Only support native password & caching sha2 password here.
-        return TRILOGY_PROTOCOL_VIOLATION;
+    if (conn->socket->opts.password_len > 0) {
+        if (strcmp("mysql_native_password", auth_switch_packet.auth_plugin) &&
+            strcmp("caching_sha2_password", auth_switch_packet.auth_plugin)) {
+            // Only support native password & caching sha2 password here.
+            return TRILOGY_PROTOCOL_VIOLATION;
+        }
     }
 
     memcpy(handshake->auth_plugin, auth_switch_packet.auth_plugin, sizeof(auth_switch_packet.auth_plugin));


composerinteralia:
Yup, all of that matches my understanding.

> It seems strange that MySQL 8.0 would change the default

There are more details about that change in https://dev.mysql.com/doc/refman/8.0/en/caching-sha2-pluggable-authentication.html and https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password

> my knowledge of SQL authentication methods is limited

Saaaame.

The fact that it's a caching plugin had me wondering whether there might be a different path based on whether it's a cache hit or miss, but based on https://dev.mysql.com/doc/dev/mysql-server/latest/page_caching_sha2_authentication_exchanges.html#sect_caching_sha2_info it seems like we should go down the same quick path every time if the password is empty. 🤷🏻‍♂️ 

>  I wonder if explicitly setting default_authentication_plugin to caching_sha2_password would help?

I would think we'd go the other way, setting it to `native_mysql_password` so that our 8.0 tests work like the 5.7 tests. That might be OK, since I don't think we really support `caching_sha2_password` yet anyway (e.g. https://github.com/github/trilogy/issues/26) except maybe for empty passwords.

> Even with the CI docker files I've yet to get the test to fail locally on my machine, which is certainly odd

Interesting!

adrianna-chang-shopify:
> I would think we'd go the other way, setting it to native_mysql_password so that our 8.0 tests work like the 5.7 tests. That might be OK, since I don't think we really support caching_sha2_password yet anyway (e.g. https://github.com/github/trilogy/issues/26) except maybe for empty passwords.

Yeah maybe that's a sensible fix in the meantime, to revert back to the 5.7 default. I didn't realize there were other ongoing issues with `caching_sha2_password`. Is anyone from Github looking into that actively?

composerinteralia:
Closing since I don't think this is the right solution. To be continued in https://github.com/github/trilogy/issues/43

PR #44, opened 2023-02-02
Have test_connection_error test user account with wrong password

adrianna-chang-shopify:
This test was being skipped for reasons described in https://github.com/github/trilogy/issues/43.
Instead of testing for a connection error by using a nonexistent user, we can use an existing user with the wrong password. This appears to avoid the authentication plugin flakiness, while still executing the code path we care about related to an "Access denied" connection error.

Is it still worth having a skipped test with an invalid username so that the issue remains on our radar?

### DIFF
--- a/contrib/ruby/test/client_test.rb
+++ b/contrib/ruby/test/client_test.rb
@@ -531,12 +531,11 @@ class ClientTest < TrilogyTest
   end
 
   def test_connection_error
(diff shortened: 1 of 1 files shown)


composerinteralia:
> Is it still worth having a skipped test with an invalid username so that the issue remains on our radar?

I don't think we need a skipped test. We can keep https://github.com/github/trilogy/issues/43 open.

PR #47, opened 2023-02-24
Use `rb_io_wait` function and cache `io` instance.

ioquatix:
https://github.com/github/activerecord-trilogy-adapter/issues/28


contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -49,12 +53,29 @@ static size_t trilogy_memsize(const void *ptr) {
     return memsize;
 }
 
+static void trilogy_ctx_mark(void *ptr)
+{
+    struct trilogy_ctx *ctx = ptr;
+
+    if (RTEST(ctx->io))
```
matthewd, line 60: Unnecessary?

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -183,6 +205,7 @@ static double timeval_to_double(struct timeval tv)
 
 static int _cb_ruby_wait(trilogy_sock_t *sock, trilogy_wait_t wait)
 {
+    struct trilogy_ctx *ctx = sock->user_data;
```
matthewd, line 208: This almost feels like there's a danger of a possible lifetime issue here, but I think it's actually fine.

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -211,10 +234,16 @@ static int _cb_ruby_wait(trilogy_sock_t *sock, trilogy_wait_t wait)
...
 
     int fd = trilogy_sock_fd(sock);
-    if (rb_wait_for_single_fd(fd, wait_flag, timeout) <= 0)
-        return TRILOGY_SYSERR;
+    if (ctx->io == Qnil) {
+        ctx->io = rb_io_fdopen(fd, 0, "trilogy_socket");
+    }
 
-    return 0;
+    VALUE result = rb_io_wait(ctx->io, RB_INT2NUM(wait_flag), rb_fiber_scheduler_make_timeout(timeout));
```
matthewd, line 241: I acknowledge it's what's already happening on sufficiently-new rubies (along with the IO construction), but `rb_fiber_scheduler_make_timeout` sounds like it might allocate. Is that true? Something we can/should avoid?

matthewd (commented):
Thanks for following up on this @ioquatix!

contrib/ruby/ext/trilogy-ruby/cext.c
ioquatix: There is a small chance I the future it might allocate but for the moment it doesn't, it just embeds a float into the VALUE. I hope at some point in the future to migrate to fixed point timeouts.

contrib/ruby/ext/trilogy-ruby/cext.c
ioquatix: I didn't look too closely at the lifetimes but I assume the socket should outlive the context?

contrib/ruby/ext/trilogy-ruby/cext.c
ioquatix: Mark moveable doesn't check if the field is NULL, IIRC without this it can crash.

matthewd: I was looking at https://github.com/ruby/ruby/blob/1fdaa0666086529b3aae2d509a2e71c4247c3a12/gc.c#L7108 and https://github.com/ruby/ruby/blob/1fdaa0666086529b3aae2d509a2e71c4247c3a12/gc.c#L10395 -- I'm not _100%_ sure it's part of their defined contract, but I imagine it must be true by ossification, even if not originally intended?

Intuitively I just prefer "this slot contains a a possible GC ref", rather than an [even trivial] encoding of "we _might_ need to mark this". 🤷🏻‍♂️

contrib/ruby/ext/trilogy-ruby/cext.c
matthewd: Bonus question: is this going to affect our min ruby version (which I note we don't seem to currently declare)? I'm guessing this is moderately recent.

ioquatix: No I'll add feature detection, this PR is just what I could manage in the 30 minutes I had before bed time.

ioquatix:
@matthewd can you please approve and run.

ioquatix:
There is one more piece we should consider exposing, which is the "resolve DNS" to use the fiber scheduler hook.

ioquatix:
So, writing this PR has made me realise we need a set of shims for the blocking C APIs, e.g. `open`, `addrinfo`, etc. We have the code, it's just not exposed. I supposed we can do that for Ruby 3.3.

We should also extend the test suite to run within a fiber scheduler (probably Async).

matthewd:
@ioquatix  is this good to go? It seems fine to me, not sure if you've kept it in draft for a reason, or just not hit the button?

ioquatix:
It's probably as good as we can make it.. but I want to review it this weekend. This week has been pretty busy, have not had much time for open source stuff.

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -211,10 +253,23 @@ static int _cb_ruby_wait(trilogy_sock_t *sock, trilogy_wait_t wait)
     }
 
     int fd = trilogy_sock_fd(sock);
+
+#ifdef TRILOGY_RB_IO_WAIT
+    if (ctx->io == Qnil) {
+        ctx->io = rb_io_fdopen(fd, O_RDWR, NULL);
```
casperisfine, line 259: `trilogy_data_type` is marked with `RUBY_TYPED_WB_PROTECTED`, so you have to trigger a write barrier here.

Problem is you don't have access to the `VALUE` here.

ioquatix: Okay, so we should pass around the `VALUE` then?

casperisfine: Either that, or create the IO eagerly in a place where you have access to it.

contrib/ruby/ext/trilogy-ruby/cext.c
eregon: > it just embeds a float into the VALUE

Flonums only work for a subset of all double, if it doesn't fit it will be an allocation.

casperisfine: I think it also always allocate on 32bits?

kirs:
👋 just wanted to say that  appreciate this WIP PR by Samuel and I'd really really love to see this happen in the upstream, and to be able to use trilogy with fiber scheduler for IO-heavy workloads.

casperisfine:
> I'd really really love to see this happen in the upstream, and to be able to use trilogy with fiber scheduler for IO-heavy workloads.

So my knowledge of the fiber scheduler is limited, but AFAIK you can already use it today? My understanding is that this PR make it a bit more efficient, but `rb_wait_for_single_fd` is already enough to be usable.

As for it happening, there's not much work needed, all the blockers are listed in review comments.

ioquatix:
I've rebased and updated the PR.

`rb_wait_for_single_fd` is okay but it's a slow path. `rb_io_wait` is the fast path.

@matthewd can you please run CI.

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -28,26 +38,59 @@ static ID id_socket, id_host, id_port, id_username, id_password, id_found_rows,
...
+    rb_gc_mark_movable(ctx->encoding);
 }
 
-static void free_trilogy(void *ptr)
+static void trilogy_ctx_compact(void *ptr)
 {
     struct trilogy_ctx *ctx = ptr;
-    trilogy_free(&ctx->conn);
+
+    rb_gc_location(ctx->self);
```
casperisfine, line 72:
```suggestion
    ctx->self = rb_gc_location(ctx->self);
```

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -28,26 +38,59 @@ static ID id_socket, id_host, id_port, id_username, id_password, id_found_rows,
...
+static void trilogy_ctx_mark(void *ptr)
 {
     struct trilogy_ctx *ctx = ptr;
-    rb_gc_mark(ctx->encoding);
+
+    rb_gc_mark_movable(ctx->self);
+
+#ifdef TRILOGY_RB_IO_WAIT
+    if (RTEST(ctx->io))
+        rb_gc_mark_movable(ctx->io);
```
casperisfine, line 62:
```suggestion
    rb_gc_mark_movable(ctx->io);
```

`gc_mark` don't mind `false/nil`

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -224,6 +274,62 @@ struct rb_trilogy_wait_args {
...
+
+    default:
+        return TRILOGY_ERR;
+    }
+
+    if (ctx->io == Qnil) {
+        RB_OBJ_WRITE(ctx->self, &ctx->io, rb_io_fdopen(trilogy_sock_fd(sock), O_RDWR, NULL));
+    }
+
+    if (timeout) fprintf(stderr, "_cb_ruby_wait timeout: %ld.%06ld\n", timeout->tv_sec, timeout->tv_usec);
```
casperisfine, line 319:
```suggestion
```

contrib/ruby/ext/trilogy-ruby/cext.c
ioquatix: Thanks I wonder where I got the idea I should check first.

ioquatix:
I recall why I gave up on this PR for a while, we were missing `rb_io_open_descriptor` which is needed to create a "temporary non-owning `FMODE_EXTERNAL` IO instance. Since that's now available in recent CRuby, it is possible to implement it correctly.

contrib/ruby/ext/trilogy-ruby/cext.c
ioquatix: Thanks this is now fixed, I added `user_data` and:

```c
    if (ctx->io == Qnil) {
        VALUE io = rb_io_open_descriptor(rb_cIO, trilogy_sock_fd(sock), FMODE_EXTERNAL, RUBY_Qnil, RUBY_Qnil, NULL);
        RB_OBJ_WRITE(ctx->self, &ctx->io, io);
    }
```

contrib/ruby/ext/trilogy-ruby/cext.c
ioquatix: Fixed.

ioquatix:
I believe all the feedback has been addressed, so this is good to go.

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -224,6 +270,66 @@ struct rb_trilogy_wait_args {
...
+    if (ctx->io == Qnil) {
+        VALUE io = rb_io_open_descriptor(rb_cIO, trilogy_sock_fd(sock), FMODE_EXTERNAL, RUBY_Qnil, RUBY_Qnil, NULL);
+        RB_OBJ_WRITE(ctx->self, &ctx->io, io);
+    }
+
+    if (timeout->tv_sec == 0 && timeout->tv_usec == 0) {
+        timeout = NULL;
+    }
+
+    VALUE result = rb_io_wait(ctx->io, RB_INT2NUM(wait_flag), rb_fiber_scheduler_make_timeout(timeout));
```
eregon, line 320: rb_fiber_scheduler_make_timeout() is not implemented yet on TruffleRuby: https://github.com/search?q=repo%3Aoracle%2Ftruffleruby%20rb_fiber_scheduler_make_timeout&type=code
Yet `#if defined(HAVE_RB_IO_WAIT) && defined(RB_IO_OPEN_DESCRIPTOR) && defined(HAVE_RUBY_FIBER_SCHEDULER_H)` should be true on TruffleRuby.

So could you check for `have_func("rb_fiber_scheduler_make_timeout")` in extconf.rb?

(BTW [that function](https://github.com/ruby/ruby/blob/97e6ad49a4604e7e4ca04da2aaafc63cbd5d29d8/scheduler.c#L270-L278) uses `0.000001f` (32-bit float) shouldn't it be `0.000001` (64-bit double) instead?)

eregon (commented):
What's the performance gain with this?
Is it worth the duplication of `_cb_ruby_wait` and additional complexity?

contrib/ruby/ext/trilogy-ruby/cext.c
ioquatix: Yes we can fix those issues, thanks for pointing them out.

ioquatix:
> What's the performance gain with this?

It should be slightly more efficient, but I imagine there isn't much performance difference (except for GC overhead due to the temporary IO objects created).

> Is it worth the duplication of _cb_ruby_wait and additional complexity?

I would say yes. `rb_wait_for_single_fd` creates a temporary IO instance for each call, which creates extra (pointless) garbage. It's better for Ruby extensions that do IO to use a proper IO object IMHO.

contrib/ruby/ext/trilogy-ruby/cext.c
ioquatix: Here is the fix to prefer double math: https://github.com/ruby/ruby/pull/12803

ioquatix:
I think now that `rb_trilogy_wait_protected` was extracted out, maybe we can simplify this PR a little to address the complexity @eregon mentioned.

contrib/ruby/ext/trilogy-ruby/cext.c
ioquatix: Okay, fixed.

PR #48, opened 2023-02-24
Expose `#connection_options` on the Ruby binding.

adrianna-chang-shopify:
This PR exposes a `#connection_options` method on the Ruby binding, allowing users to query the client for the options passed in at connect time. This is meant to be similar to [Mysql2's `#query_options`](https://www.rubydoc.info/gems/mysql2/Mysql2/Client#query_options-instance_method).

We're currently using `Mysql2::Client#query_options`  in a couple of places in our Rails apps at Shopify, and there's no equivalent in Trilogy -- we could go through the connection pool and grab the `db_config` instead, but it's nice to be able to access connection options on the client IMHO. It also facilitates other use cases, like being able to configure span name for MySQL instrumentation with OpenTelemetry (just as is being done for [Mysql2](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb#L117-L120)).

As an aside: I've noticed that there are a couple of places (e.g. benchmark [script](https://github.com/github/trilogy/blob/main/contrib/ruby/script/benchmark#L86), [docs](https://github.com/github/trilogy/tree/main/contrib/ruby#mysql2-gem-compatibility)) where we reference a non-existent `#query_options` method. I'm wondering if this was removed from the API at some point? Is there any value to bringing that back in addition to `#connection_options`? If not, I can remove those references to avoid confusion 😄

### DIFF
--- a/contrib/ruby/lib/trilogy.rb
+++ b/contrib/ruby/lib/trilogy.rb
@@ -84,6 +84,10 @@ class Trilogy
     include ConnectionError
   end
 
+  def connection_options
+    @connection_options.dup.freeze
+  end
+
   def in_transaction?
     (server_status & SERVER_STATUS_IN_TRANS) != 0
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 
 ### Added
-  - Implement multi-statement support on the Ruby binding. 
+  - Implement multi-statement support on the Ruby binding.
+  - Add `#connection_options` method to Ruby binding.
 
 ## 2.2.0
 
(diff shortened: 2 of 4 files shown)


composerinteralia:
I'll try to take a look at this PR early next week.

> As an aside: I've noticed that there are a couple of places (e.g. benchmark [script](https://github.com/github/trilogy/blob/main/contrib/ruby/script/benchmark#L86), [docs](https://github.com/github/trilogy/tree/main/contrib/ruby#mysql2-gem-compatibility)) where we reference a non-existent #query_options method. I'm wondering if this was removed from the API at some point? Is there any value to bringing that back in addition to #connection_options? If not, I can remove those references to avoid confusion 😄

It looks like `query_options` got replaced with `query_flags`, replacing some `rb_hash_aref` in `load_query_options` with bit checking. From the original commit message for that change:

> Use a bitfield to store the query flags instead of in a Ruby hash. This
will reduce allocations and make queries faster.

adrianna-chang-shopify:
> It looks like query_options got replaced with query_flags

Ah, thanks Daniel! I was wondering if that was maybe the case. I can cut a quick PR to update those references.

contrib/ruby/lib/trilogy.rb
```diff
@@ -1,6 +1,8 @@
 require "trilogy/version"
 
 class Trilogy
+  attr_reader :connection_options
```
composerinteralia, line 4: Do think it's worth dup-ing and freezing on read to limit folks mucking around with the original hash passed in?

```suggestion
  def connection_options
    @connection_options.dup.freeze
  end
```

I suppose folks could still muck around with the hash values, but that seems somewhat less likely.

composerinteralia (approved):
I'm fine with this change if it's helpful for you 👍🏻, but had one thought about discouraging folks from modifying the options via Trilogy.

contrib/ruby/lib/trilogy.rb
adrianna-chang-shopify: Makes sense to me 👍

PR #49, opened 2023-02-28
Preserve DatabaseError constant

composerinteralia:
https://github.com/github/trilogy/pull/15 changed the error class for TRILOGY_ERR from `DatabaseError` to `ProtocolError`. This commit reintroduces `DatabaseError` as an ancestor of `ProtocolError` for compatibility reasons, so that rescuing `DatabaseError` still works (mostly) as before.

Note that even with this change there are a number of error codes (see `ProtocolError::ERROR_CODES`) that are no longer `DatabaseError`s, since we have explicitly recategorized them as `BaseConnectionError`, `TimeoutError`, or `QueryError`.

### DIFF
--- a/contrib/ruby/lib/trilogy.rb
+++ b/contrib/ruby/lib/trilogy.rb
@@ -51,7 +51,15 @@ class Trilogy
     end
   end
 
+  # DatabaseError was replaced by ProtocolError, but we'll keep it around as an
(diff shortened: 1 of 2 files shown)


composerinteralia:
This is enough for me to be comfortable with a minor version bump. But if we wanted to be strict about it we could add new subclasses of `BaseConnectionError`, `TimeoutError`, and `QueryError` that `include DatabaseError` and then use those in `ProtocolError::ERROR_CODES` instead.

PR #53, opened 2023-03-07
Raise system error subclasses

composerinteralia:
https://github.com/github/trilogy/pull/41 added a translation of `ECONNREFUSED` and `ECONNRESET` to `Trilogy::BaseConnectionError`.

We've got a few places at GitHub where we explicitly rescue these errors (or `SystemCallError`). Although it is possible to rework those rescues to check for particular error messages (like we did in
https://github.com/github/activerecord-trilogy-adapter/commit/03c16105c8ea5da4c90f837b607e97b1b2dcd6d0), it'd be a bit easier if we could maintain compatibility. I also worry that checking for specific error messages could be a bit brittle.

This commit introduces two new Trilogy errors, which inherit from the corresponding `Errno` errors. We've already done something like this for `Errno::ETIMEDOUT`.

This keeps Trilogy compatible with GitHub's existing rescues, while still raising something that is a `Trilogy::Error` and a `Trilogy::ConnectionError` (although NOT a
`Trilogy::BaseConnectionError` anymore).

A downside to this approach is that we might end up with a lot of Trilogy-specific `Errno` subclasses. If that gets annoying perhaps we could revisit when it comes time for a major release. Another approach might be a single `Trilogy::SystemCallError` that inherits from `SystemCallError` and then keeps a reference to the underlying `Errno` error.

cc @adrianna-chang-shopify

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -16,7 +16,8 @@
 
 VALUE Trilogy_CastError;
 static VALUE Trilogy_BaseConnectionError, Trilogy_ProtocolError, Trilogy_SSLError, Trilogy_QueryError,
-    Trilogy_ConnectionClosedError, Trilogy_TimeoutError, Trilogy_Result;
+    Trilogy_ConnectionClosedError, Trilogy_ConnectionRefusedError, Trilogy_ConnectionResetError,
+    Trilogy_TimeoutError, Trilogy_Result;
 
 static ID id_socket, id_host, id_port, id_username, id_password, id_found_rows, id_connect_timeout, id_read_timeout,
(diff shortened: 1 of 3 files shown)


adrianna-chang-shopify (commented):
Makes sense to me!

As a heads up, I was planning to submit a PR in the next week or so with proposed breaking changes to the error classes (ie. moving away entirely from the syscall errors) in anticipation of Trilogy being upstreamed. Makes sense to have these backwards compatible changes shipped ahead of that major version bump though 👍 Is your team still on board with those changes taking place?

composerinteralia:
Yeah, I think it's fine to make further breaking changes. If the changes here are going in the wrong direction, I can try something else. The main thing is that we'd like to rescue these separately from other Trilogy errors, so `Trilogy_BaseConnectionError` isn't ideal.

adrianna-chang-shopify:
Maintaining the parallel to the way we've handled `Errno::ETIMEDOUT` SGTM 👍

PR #57, opened 2023-03-13
Support `TRILOGY_CAPABILITIES_MULTI_RESULTS`

adrianna-chang-shopify:
### Proposed Changes
The MySQL client library enables `CLIENT_MULTI_RESULTS` either by passing the CLIENT_MULTI_RESULTS flag itself, or implicitly by passing `CLIENT_MULTI_STATEMENTS` (which also enables `CLIENT_MULTI_RESULTS`). `CLIENT_MULTI_RESULTS` is enabled by default.

Consequently, we should set the `TRILOGY_CAPABILITIES_MULTI_RESULTS` client flag along with `TRILOGY_CAPABILITIES_MULTI_STATEMENTS` when the Trilogy client is initialized with `multi_statement: true`. 

Since the client flag can be set explicitly as well, this PR allows `TRILOGY_CAPABILITIES_MULTI_RESULTS` to be set independently via `multi_result: true`. This enables multi-result processing without allowing the client to send multi-statement queries, which is useful for stored procedures, dynamic SQL in init files, etc.

To match `Mysql2`'s behaviour, we enable `TRILOGY_CAPABILITIES_MULTI_RESULTS` by default. There shouldn't be any downsides to enabling this flag; it essentially specifies that the client is capable of multi-result set processing, which it is thanks to Paarth's changes in #35.

### Motivation for Change
At Shopify, this change is important for us as we attempt to migrate from Mysql2 to Trilogy, because ProxySQL treats the two drivers differently if they have different client flags set. This results in a lot of `COM_CHANGE_USER` commands if the two clients are running simultaneously. 

[This](https://github.com/sysown/proxysql/blob/v2.x/lib/mysql_connection.cpp#L668-L681) is where ProxySQL compares the client flags and [this](https://github.com/sysown/proxysql/blob/75a78895ce6b3f6e71ed44fa6308158e36d1de3d/lib/MySQL_Thread.cpp#L5362) is where ProxySQL determines whether it can use the cached connections, or whether it needs to perform a `COM_CHANGE_USER` command. The issue for us was that running both of the adapters at 50% resulted in a ton of `COM_CHANGE_USER` commands as the connection pool was handed off between these two users, and we overloaded MySQL.

We could, of course, just configure the flag ourselves, but it makes sense to me to match `Mysql2`'s behaviour.

cc @byroot @paarthmadan @eileencodes


casperisfine (commented):
Patch looks great 👍

eileencodes (approved):
Looks good to me

cc/ @matthewd @HParker @jhawthorn @composerinteralia

PR #60, opened 2023-03-18
Add a query flag to toggle the type of casting done by SUM

nickborromeo:
In https://github.com/github/trilogy/pull/37 there was a change that updated the value of SUM to be Integer if what is being summed are all Integers. It used to be that it will _always_ return a Decimal type.

When we pulled in that change, there were tests that failed because we were doing arthimitic operations on the result of queries that have SUM. Ruby will have different results if you divide intergers and decimals.

In order to keep that behavior for us (GitHub) while allowing other consumers to use SUM and expect the _right_ type we are introducing a new query flag that we can set in our adapter. This query flag will be what will determine if we should be casting the value of SUM to follow the _new_ behavior or roll back to the _old_ behavior

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/trilogy-ruby.h
+++ b/contrib/ruby/ext/trilogy-ruby/trilogy-ruby.h
@@ -9,6 +9,7 @@
 #define TRILOGY_FLAGS_CAST_BOOLEANS 2
 #define TRILOGY_FLAGS_LOCAL_TIMEZONE 4
 #define TRILOGY_FLAGS_FLATTEN_ROWS 8
+#define TRILOGY_FLAGS_CAST_SUM 16
 #define TRILOGY_FLAGS_DEFAULT (TRILOGY_FLAGS_CAST)
 
 struct rb_trilogy_cast_options {
@@ -16,6 +17,7 @@ struct rb_trilogy_cast_options {
     bool cast_booleans;
     bool database_local_time;
     bool flatten_rows;
+    bool cast_sum;
 };
 
 struct column_info {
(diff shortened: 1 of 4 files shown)


composerinteralia (commented):
The code looks good to me. I think this applies to more than just sum (I wasn't thinking about that as we were browsing sum queries earlier today), so we may need a more general name, but I don't have any good ideas at the moment. Maybe something will come to us over the weekend 😄

nickborromeo:
>  I think this applies to more than just sum

ah good to know, if that is the case we can generalize it and may use aggregate since that is what SQL buckets those functions in anyway. Maybe something like `TRILOGY_FLAGS_CAST_AGGREGATES` or `TRILOGY_FLAGS_CAST_AGGREGATIONS`

contrib/ruby/ext/trilogy-ruby/trilogy-ruby.h
```diff
@@ -9,13 +9,15 @@
 #define TRILOGY_FLAGS_CAST_BOOLEANS 2
 #define TRILOGY_FLAGS_LOCAL_TIMEZONE 4
 #define TRILOGY_FLAGS_FLATTEN_ROWS 8
+#define TRILOGY_FLAGS_CAST_SUM 16
```
composerinteralia, line 12: I'm not 100% certain this is limited to aggregate functions either. This is a bit verbose, but what do you think about something like:

```suggestion
#define TRILOGY_FLAGS_CAST_ALL_DECIMALS_TO_BIGDECIMAL 16
```

composerinteralia (approved):
👍🏻

PR #62, opened 2023-03-23
Fix CHANGELOG

adrianna-chang-shopify:
I messed up the PR num on my last commit. Also added an entry for the new TRILOGY_FLAGS_CAST_ALL_DECIMALS_TO_BIGDECIMALS flag.


composerinteralia (approved):
Thanks for keeping this up to date! Makes releasing new versions much easier.

PR #63, opened 2023-03-27
Raise all SystemCallErrors as `Trilogy::Error`

adrianna-chang-shopify:
### Background

See https://github.com/github/trilogy/pull/58 for more context.

This allows syscall errors to be rescued generically by `Trilogy::Error`, or handled individually by consumers using their original Errno classes.

To make this work, we subclass all of the SystemCallErrors and include the base error module on each of them.

I've also removed the constructors from the existing Errno class subclasses (`Trilogy::TimeoutError`, `Trilogy::ConnectionClosedError`, etc.). We never set an `error_code` on syscall errors, so the constructors shouldn't be needed.

If anyone has input on a better way to handle including the `Trilogy::Error` module on all of the Errno classes, let me know! Enumerating descendants of `SystemCallError`, subclassing them, and including `Trilogy::Error` seems like the most straightforward approach so far, but is admittedly not pretty.

### DIFF
--- a/contrib/ruby/lib/trilogy.rb
+++ b/contrib/ruby/lib/trilogy.rb
@@ -15,6 +15,27 @@ class Trilogy
     include Error
   end
 
+  # Trilogy may raise various syscall errors, which we treat as Trilogy::Errors.
+  class SyscallError
+    ERRORS = {}
+
+    Errno.constants
+      .map { |c| Errno.const_get(c) }.uniq
+      .select { |c| c.is_a?(Class) && c < SystemCallError }
+      .each do |c|
+        errno_name = c.to_s.split('::').last
+        ERRORS[c::Errno] = const_set(errno_name, Class.new(c) { include Trilogy::Error })
+      end
+
+    ERRORS.freeze
+
+    class << self
+      def from_errno(errno, message)
+        ERRORS[errno].new(message)
+      end
+    end
+  end
+
   class BaseError < StandardError
     include Error
 
@@ -43,29 +64,14 @@ class Trilogy
 
   class TimeoutError < Errno::ETIMEDOUT
     include ConnectionError
-
-    def initialize(error_message = nil, error_code = nil)
-      super
-      @error_code = error_code
-    end
   end
 
   class ConnectionRefusedError < Errno::ECONNREFUSED
     include ConnectionError
-
-    def initialize(error_message = nil, error_code = nil)
-      super
-      @error_code = error_code
-    end
   end
 
   class ConnectionResetError < Errno::ECONNRESET
     include ConnectionError
-
-    def initialize(error_message = nil, error_code = nil)
-      super
-      @error_code = error_code
-    end
   end
 
   # DatabaseError was replaced by ProtocolError, but we'll keep it around as an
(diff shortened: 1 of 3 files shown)


matthewd:
Yeah my concern is that it's really hard to define a list of plausibly-expectable errors (e.g. EHOSTUNREACH probably wouldn't make an initial cut, but considered properly does seem like a thing that an interested caller could reasonably want to specifically handle as a connection issue -> retry / whatever).

Not to mention the theoretical variation across platforms once we stray from the basics.


My first thought was to use a lazy hash... which doesn't feel like a particularly good idea, and would be messy though maybe not ridiculous. It's a bounded list, and most processes most of the time won't ever see / need most classes.

This feels terrible but I can't see an alternative that doesn't pre-encode a subset of errnos that are "allowed" to happen. :confused:

contrib/ruby/lib/trilogy.rb
```diff
@@ -13,6 +13,24 @@ module ConnectionError
...
+
+    Errno.constants
+      .map { |c| Errno.const_get(c) }
+      .select { |c| c.is_a?(Class) && c < SystemCallError }
+      .each { |c| ERRORS[c::Errno] = Class.new(c) { include Trilogy::Error } }
+
+    ERRORS.freeze
+
+    class << self
+      def from_errno(message, errno)
```
matthewd, line 28: Nit: feels to me like `from_errno`'s first argument should be `errno`

adrianna-chang-shopify:
Oh actually lazy hash sounds like a good idea? e.g. something like this right?
```ruby
  class SyscallError
    ERRORS = {}
    class << self
      def from_errno(errno, message)
        ERRORS.fetch(errno) do |errno|
          errno_name = Errno.constants.find { |c| Errno.const_get(c)::Errno == errno }
          errno_subclass = Class.new(Errno.const_get(errno_name)) { include Trilogy::Error }
          ERRORS[errno] = const_set(errno_name, errno_subclass)
        end.new(message)
      end
    end
  end
```

That way we avoid subclassing the unlikely ones.

adrianna-chang-shopify:
> Until very recently (3.2) that bumps the global constant cache

Oh good to know, thank you Jean 🙇‍♀️ 

> That said, defining the 150-ish classes would use about 30kB

I think it would be closer to 20kbB; several of the `Errno` constants map to `Errno::NOERROR`. With `#uniq`:
```
>> Errno.constants.map { |c| Errno.const_get(c) }.uniq.select { |c| c.is_a?(Class) && c < SystemCallError }.size
=> 107
>> ObjectSpace.memsize_of(Class.new(StandardError)) * 107
=> 20544
```

adrianna-chang-shopify:
@byroot @matthewd do we feel comfortable shipping this in its current form?

byroot:
Fine by me.

matthewd:
I think @casperisfine was advising against the combo of `const_set` + deferred subclassing, not necessarily `const_set` + immediate-subclass loop.

Failing that, I think we'd want to define both `self.inspect` and `inspect`:

```
Class.new(Errno::EBADF).new
=> #<#<Class:0x0000000104db7e50>: Bad file descriptor>
```

byroot:
> not necessarily `const_set` + immediate-subclass loop.

Yeah, if it's eagerly defined it's fine.

adrianna-chang-shopify:
Ahh gotcha, yeah that makes more sense in terms of invalidating the global constant cache. Let me bring back back the const_set then 👍

matthewd (approved):
:shipit:

PR #66, opened 2023-04-10
[Docs] Fix Trilogy `TRILOGY_CAPABILITIES` docs

adrianna-chang-shopify:
This PR tweaks a couple of things with our `TRILOGY_CAPABILITIES` docs: 

* I don't think the description for `TRILOGY_CAPABILITIES_TRANSACTIONS` is correct -- AFAICT it _is_ supported, so I've updated the description to match.
* Same thing with `TRILOGY_CAPABILITIES_SSL` - it said this flag wasn't supported, but Trilogy has SSL support.
* We were also saying that `MULTI_STATEMENTS` still wasn't supported, so removed that line too.

### DIFF
--- a/inc/trilogy/protocol.h
+++ b/inc/trilogy/protocol.h
@@ -90,16 +90,14 @@
      * From client: the client is interactive.                                                                         \
      */                                                                                                                \
     XX(TRILOGY_CAPABILITIES_INTERACTIVE, 0x00000400)                                                                   \
-    /* Not implemented.                                                                                                \
-     *                                                                                                                 \
-     * From server: the server supports ssl.                                                                           \
+    /* From server: the server supports ssl.                                                                           \
      *                                                                                                                 \
      * From client: tells the server it should switch to an ssl connection.                                            \
      */                                                                                                                \
     XX(TRILOGY_CAPABILITIES_SSL, 0x00000800)                                                                           \
-    /* Not used.                                                                                                       \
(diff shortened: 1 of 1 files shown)


inc/trilogy/protocol.h
```diff
@@ -10,10 +10,10 @@
 #include <string.h>
 
 #define TRILOGY_CAPABILITIES(XX)                                                                                       \
-    XX(TRILOGY_CAPABILITIES_NONE, 0)                                                                                   \
     /* Not used. This flag is assumed by current servers.                                                              \
```
jhawthorn, line 13: This seems a bit odd describing `TRILOGY_CAPABILITIES_NONE` (which is the absence of capabilities rather than a flag). Maybe this should be removed?

jhawthorn:
One suggestions otherwise 👍

inc/trilogy/protocol.h
adrianna-chang-shopify: Yep, sorry that's my bad. I read "assumed by current servers" and thought it was describing this as "the absence of capabilities assumed by current servers as the default", and thought maybe it was intended to describe `TRILOGY_CAPABILITIES_NONE` instead of `TRILOGY_CAPABILITIES_LONG_PASSWORD` below. I'm realizing that doesn't really make sense 🤦‍♀️

jhawthorn (approved):

PR #68, opened 2023-04-17
No need to accept `multi_result` as conn option in Ruby binding

adrianna-chang-shopify:
Follow up to https://github.com/github/trilogy/pull/57

We added the ability to pass `:multi_result` as a connection option to the client in the Ruby binding, but the `MULTI_RESULTS` flag is set by default so this isn't necessary. (I made the changes to the Ruby binding before we decided to set the flag as part of the default bitmask).

Might as well clean this code up in the C-ext :)


paarthmadan (commented):
Makes sense to me, always better to reduce the number of options, especially if this is a default (and expected to always be enabled)

eileencodes (approved):

PR #69, opened 2023-04-19
Set `error_code` on Trilogy::TimeoutError

adrianna-chang-shopify:
I mistakenly removed the constructor that sets `error_code` on `Trilogy::TimeoutError` in https://github.com/github/trilogy/pull/63. I thought that since it was a syscall error, we'd never set an error code on timeout errs, but we _do_ raise a `TimeoutError` with an error code [here](https://github.com/github/trilogy/blob/main/contrib/ruby/lib/trilogy.rb#L87).


eileencodes (approved):
Confirmed this fixes the issue on the Rails side.

PR #70, opened 2023-04-25
Just use raw close() for discard

jhawthorn:
Follow up to #65 cc @casperisfine

Previously we were doing a dance with file descriptors in order to replace the connection's fd with one open to `/dev/null`. Instead we should be able to just `close()` the fd so long as we don't perform any writes on it.

### DIFF
--- a/contrib/ruby/test/client_test.rb
+++ b/contrib/ruby/test/client_test.rb
@@ -903,6 +903,18 @@ class ClientTest < TrilogyTest
     assert_match "TRILOGY_CLOSED_CONNECTION", error.message
   end
 
+  def test_discard_closes_connection
+    client = new_tcp_client
+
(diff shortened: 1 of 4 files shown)


casperisfine:
> so long as we don't perform any writes on it.

But that's actually the point. To make sure we won't flush any buffers etc.

Is this dance causing issues?

jhawthorn:
In practice most of the time this probably works fine, but it adds some dangers and unnecessary complexity. `open` could fail because of fd exhaustion. Calling `SSL_shutdown` may push errors (which we should be cleaning up anyways). It all _should_ work, I'd just like something similar and not to have an extra state to consider connections could be in (being connected to `/dev/null`)

I've opened #111 which simplifies this even further and combines the shutdown and discard callbacks.

PR #73, opened 2023-05-05
Fix build with LibreSSL

jeremyevans:
Using ifdef here seems preferable to checking version numbers anyway.


composerinteralia:
Makes sense to me. Thanks! (Pardon the failing stretch tests. Our CI needs a bit of a refresh.)

PR #74, opened 2023-05-17
 validate max packet size

kamil-gwozdz:
This PR achieves the same goal as https://github.com/rails/rails/pull/48226 intended to but it's done in the client layer.

### Motivation:
Trilogy doesn't do anything to protect against sending queries bigger than the max packet size. Depending on timing and query size, sending a giant query (the default max is 64MB…) could result in either Trilogy::Error: trilogy_query_send: TRILOGY_CLOSED_CONNECTION, Errno::ECONNRESET: Connection reset by peer - trilogy_query_send, or Trilogy::DatabaseError: trilogy_query_recv: 1153 Got a packet bigger than 'max_allowed_packet' bytes.

### Details
This PR just shifts the error to the client side so it can be raised without attempting to sent the query, which used to result in aborting the connection.

It will also improve visibility for this kind of error. Before, we weren't always getting the corresponding error message because sometimes the connection was aborted before we received it from the server.

---
merge after https://github.com/github/trilogy/pull/78

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -771,11 +771,12 @@ static VALUE read_query_response(VALUE vargs)
 }
 
 static void validate_query_size(VALUE self, VALUE query, ID id_max_allowed_packet) {
-    char *c_query = StringValueCStr(query);
     VALUE max_allowed_packet = rb_ivar_get(self, id_max_allowed_packet);
+    char *c_query = StringValueCStr(query);
+    unsigned long c_query_size = strlen(c_query);
 
-    if (max_allowed_packet != Qnil && strlen(c_query) >= NUM2ULONG(max_allowed_packet)) {
-        rb_raise(Trilogy_QueryError, "Query is too big %lu vs %lu. Consider increasing max_allowed_packet.", strlen(c_query), NUM2ULONG(max_allowed_packet));
+    if (max_allowed_packet != Qnil && c_query_size >= NUM2ULONG(max_allowed_packet)) {
+        rb_raise(Trilogy_QueryError, "Query is too big %lu vs %lu. Consider increasing max_allowed_packet.", c_query_size, NUM2ULONG(max_allowed_packet));
     }
 }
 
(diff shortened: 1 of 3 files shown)


contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -1135,6 +1148,7 @@ RUBY_FUNC_EXPORTED void Init_cext()
     id_ivar_rows = rb_intern("@rows");
     id_ivar_query_time = rb_intern("@query_time");
     id_connection_options = rb_intern("@connection_options");
+    id_max_allowed_packet = rb_intern("@max_allowed_packet");
```
kamil-gwozdz, line 1151: I wasn't sure if this should live in its own instance variable or in the `@connection_options`. It

contrib/ruby/lib/trilogy.rb
```diff
@@ -172,6 +172,7 @@ def initialize(options = {})
     encoding = Encoding.find(rb_encoding)
     charset = charset_for_mysql_encoding(mysql_encoding)
     _initialize(encoding, charset, **options)
+    self.max_allowed_packet = options[:max_allowed_packet] || query_with_flags("select @@max_allowed_packet", query_flags | QUERY_FLAGS_FLATTEN_ROWS).rows.first
```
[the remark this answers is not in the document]
kamil-gwozdz: I guess [that's](https://github.com/github/trilogy/issues/72#issuecomment-1535396774) a no

contrib/ruby/ext/trilogy-ruby/cext.c
composerinteralia: Ah, now I follow the question you asked me before. If you are passing this in to `Trilogy.new` then I think it's in `@connection_options` already, so I'd prefer to use it there and avoid the extra instance variable.

But I guess it's a bit awkward because unlike the other connection_options we're not actually using this value to connect 🤔. Does that make this a `query_flag` instead?

composerinteralia (commented):
I like the idea of catching this earlier. It avoids needless queries that we know will fail, and offers a more helpful message than the connection closed or reset errors.

That said, I'd like to get advice from somebody a bit more familiar than all of this than I. (I trust Matthew to know more than I do 😄 , but failing that maybe some of our database friends.)

I'm also not sure we've quite landed on where exactly to put this config. It's a bit different than anything else we currently have.

kamil-gwozdz:
I talked to @matthewd and we decided to put this logic in the c library and not in the ruby extension. I'll close this PR and open a new one next week.

PR #77, opened 2023-05-17
Allow disabling multi_results capability

luanzeba:
https://github.com/github/trilogy/pull/57 set multi-results capability as a default.

To safely upgrade to the latest version of Trilogy, we want to turn that option off and then turn it on one cluster at a time.

This commit removes the multi-results capability default from `TRILOGY_CAPABILITIES_CLIENT` but will still set it unless the client is initialized with `multi_results: false`.
That means existing users don't need to modify how they initialize the client when upgrading.

We used to have an option to enable multi results, but removed it in https://github.com/github/trilogy/pull/68 because the option was always a no-op. This PR adds the option back, with some modification.


composerinteralia (approved):
LGTM!

cc @adrianna-chang-shopify I think this should be a no-op for you since we still default to having the flag set, but I don't want you to be surprised by this change. We're not going to change this flag in the immediate future, so this PR will help us get back on Trilogy main in the meantime.

adrianna-chang-shopify (approved):
Thanks for the heads up @composerinteralia , yep this should be fine for us 👍

PR #78, opened 2023-05-18
refactor ci

kamil-gwozdz:
This PR:
* removes Debian stretch from the CI
* extends matrix with multiple, configurable ruby versions. before we're testing with whatever `apt-get install` gave us
* fixes a bug that caused jobs to be run twice for each PR push. before we were running [24](https://github.com/github/trilogy/pull/75/checks) checks for each push because `on.pull_reqest` already triggers the CI on push event. I limited `on.push` to the `main` branch and release tags.
* closes https://github.com/github/ruby-architecture/issues/753


docker-compose.yml
```diff
@@ -16,11 +16,15 @@ services:
...
+    image: ghcr.io/github/trilogy/ci-app:distro-${DISTRIBUTION_SLUG}-ruby-${RUBY_VERSION}-mysql-${MYSQL_VERSION}
     privileged: true
     build:
       context: .
-      dockerfile: "${DOCKERFILE}"
+      args:
+        - DISTRIBUTION=${DISTRIBUTION}
+        - RUBY_VERSION=${RUBY_VERSION}
+      cache_from: 
+        - ghcr.io/github/trilogy/ci-app:distro-${DISTRIBUTION_SLUG}-ruby-${RUBY_VERSION}-mysql-${MYSQL_VERSION}
```
kamil-gwozdz, line 27: do we need to worry about deleting those images every now and then?

.github/workflows/ci.yml
```diff
@@ -9,11 +15,20 @@ jobs:
       fail-fast: false
       matrix:
         mysql: ["5.7", "8"]
-        dockerfile: ["Dockerfile.test.buster", "Dockerfile.test.stretch", "Dockerfile.test.focal", "Dockerfile.test.bionic"]
+        distribution: ["debian:buster", "ubuntu:focal", "ubuntu:bionic"]
+        include:
+          - ruby: "2.8"
+          - ruby: "3.2"
+          - ruby: "1.9"
```
kamil-gwozdz, line 22: those values are quite random, I open for suggestions what should we put here

joshmgross, line 22: I don't think will have the result you'd expect. For every given matrix leg, this will have Ruby 1.9 since that will overwrite the previous combinations.
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations

If you wanted to run each version of Ruby agains that previous combinations, you'd want:

```suggestion
        ruby: ["2.8", "3.2", "1.9"]
```

That would be 18 different jobs as opposed to the 6 you're currently getting - https://github.com/github/trilogy/actions/runs/5024783568/jobs/9010898365

kamil-gwozdz: I don't want to run all possible combinations. I only care about all combinations of mysql x distro. And I also want to test all listed ruby versions, but I don't care about what mysql and distro they will get.

kamil-gwozdz: >  For every given matrix leg, this will have Ruby 1.9 since that will overwrite the previous combinations.

ahh I see what you mean now :/ 

I changed it to the way you suggested for now. what would be an alternative? Selecting a different distro x mysql pairs for each ruby version?

```yaml
include:
- distro: ubuntu
  mysql: 5
  ruby: 1.9
- distro: debian
  mysql: 5
  ruby: 2.8
```

joshmgross: You might want to add a `ruby` section to the matrix to default to 3.2:

```yaml
      matrix:
        mysql: ["5.7", "8"]
        distribution: ["debian:buster", "ubuntu:focal", "ubuntu:bionic"]
        ruby: ["3.2"]
        include:
          - mysql: "5.7"
            distribution: "debian:buster"
            ruby: "1.9"
          - mysql: "5.7"
            distribution: "debian:buster"
            ruby: "2.8"
```

This way Ruby `3.2` is used for every leg besides the two you're adding, and since it's in the original matrix it won't get overridden by anything in the `include`

kamil-gwozdz: thx, that's exactly what I needed!

matthewd (approved):

PR #81, opened 2023-05-26
wip

kamil-gwozdz:
a second take on https://github.com/github/trilogy/pull/79

PR #84, opened 2023-06-06
validate max_allowed_packet on the client side

matthewd:
This is another take on #79, incorporating lessons learned while working on #81.


inc/trilogy/builder.h
```diff
@@ -155,7 +173,23 @@ int trilogy_builder_write_lenenc_buffer(trilogy_builder_t *builder, const void *
...
 
+/* trilogy_builder_set_max_packet_length - Set the maximum packet length for
+ * the builder. Writing data to the builder that would cause the packet length
+ * to exceed this value will cause the builder to error.
+ *
+ * builder    - A pre-initialized trilogy_builder_t pointer
+ * max_length - The new maximum packet length to set
+ *
+ * Return values:
+ *   TRILOGY_OK     - The maximum packet length was set.
```
brianmario, line 189: Very minor nit that you don't really have to do, but would you mind indenting from `-` to line up with the one below, like the rest of the comments? I never used to do stuff like this, but `gofmt` changed my life for better or worse 😆 

I guess the rest of the comments above where `TRILOGY_MAX_PACKET_EXCEEDED` was added could use it as well so everything lines up nicely. But again, no big deal if that's annoying ;)

Makefile
```diff
@@ -12,7 +12,10 @@ EXAMPLES = example/trilogy_query
 
 UNAME_S := $(shell uname -s)
 
-ifneq ($(UNAME_S), Darwin)
+ifeq ($(UNAME_S), Darwin)
+	CFLAGS += "-I$(shell brew --prefix openssl@1.1)/include"
+	LDFLAGS += "-L$(shell brew --prefix openssl@1.1)/lib"
+else
```
brianmario, line 18: In a perfect world this could be in its own pull, but I was getting sick of adding these myself locally when compiling so I'm personally all good with it going in with this.

brianmario (commented):
Nice and clean

inc/trilogy/builder.h
matthewd: Oh, yeah, this is a bit of an ugly floater. I deliberately didn't want to give up that much horizontal space by aligning everything to the new very-long name, but now this line is only aligned with the separate preceding blocks. :confused:

I see I was also needlessly careless on the wrapped indentation below -- I copied across the misalignment from `buffer.h`, without noticing this is the first instance in this file.

I think we just need to do a bit of a general sweep... I think the alignment was a lot more consistent before one of the renames.

matthewd:
@brianmario thanks for checking this out! While I've got you -- what do you think of `s/TRILOGY_MAX_PACKET_LEN/TRILOGY_MAX_FRAGMENT_LEN/` or something?

Per the discussion in #79 (and my repeated confusion in figuring this out 😅), while those chunks certainly meet a reasonable definition of a packet, I feel like we might be better off using a different term internally. From a quick search, I think even other instances within the codebase are already using "packet" to refer to the potentially-multi-part total message.

test/builder_test.c
```diff
@@ -66,6 +66,67 @@ TEST test_builder_write_uint8_split_packet()
...
+    ASSERT_OK(err);
+
+    err = trilogy_builder_write_uint8(&builder, 0x02);
+    ASSERT_OK(err);
+
+    err = trilogy_builder_write_uint8(&builder, 0x03);
+    ASSERT_EQ(TRILOGY_MAX_PACKET_EXCEEDED, err);
+
+    trilogy_buffer_free(&buff);
+    PASS();
```
kamil-gwozdz, line 127:
```suggestion
    free(bytes);

    PASS();
```

test/builder_test.c
```diff
@@ -66,6 +66,69 @@ TEST test_builder_write_uint8_split_packet()
...
+    trilogy_builder_t builder;
+    err = trilogy_builder_init(&builder, &buff, 0);
+    ASSERT_OK(err);
+
+    size_t max = TRILOGY_MAX_PACKET_LEN * 2;
+    err = trilogy_builder_set_max_packet_length(&builder, max);
+    ASSERT_OK(err);
+
+    size_t len = max - 3;
+    uint8_t *bytes = malloc(len);
```
jhawthorn, line 112: Can we use `calloc` or similar to avoid reading uninitialized values. It doesn't really matter since this is a test, but fixing that should allow us to run under valgrind or sanitizers.

inc/trilogy/builder.h
matthewd: I fixed this one, and tidied the wrapping on the others, but couldn't bring myself to drag the other descriptions over that far. 🤷🏻‍♂️

test/builder_test.c
matthewd: Done 👍🏻 (including the existing ones)

PR #94, opened 2023-06-30
Avoid unnecessary splat in initialize

composerinteralia:
The splat here isn't necessary, and ends up creating an extra hash along the way.

Removing the splat also gets `Trilogy.new` with no arguments working again, which can me handy for local development.


matthewd (approved):
I think `initialize` should probably take a kwsplat, really... but agree this internal one is unnecessary.

PR #95, opened 2023-06-30
Don't call SSL_shutdown if the connection has not been fully established

bensheldon:
During network problems we have seen a uptick in `Trilogy::SSLError: trilogy_query_recv: SSL Error: shutdown while in init` errors.

It seems like a solution is to ensure that the shutdown only happens after the connection has been fully established: https://github.com/openssl/openssl/issues/710


matthewd (approved):
I've definitely seen this happen occasionally in the gem's tests, too. Nice!

PR #97, opened 2023-07-06
Fix casting of the time type

dbussink:
A time of 00:00:00 is perfectly valid, so we should not cast that to a nil. That results in incorrect values.


matthewd (approved):
Hey @dbussink!

Yeah, this looks like it was just an over-extrapolation from the datetime casting, where all-zero isn't so reasonable.

dbussink:
@matthewd 👋. That also would be my guess. Maybe it was even me who wrote this bug originally 😉. 

Do you think enough fixes have accumulated to warrant a new release?

PR #98, opened 2023-07-10
Handle EADDRNOTAVAIL as a BaseConnectionError instead of SyscallError

dhruvCW:
instead of letting EADDRNOTAVAIL be handled as a generic SyscallError convert it to a BaseConnectionError instead, in this case we treat it as a a TRILOGY_DNS_ERROR so that the activerecord adapters handle it as a DBConnectionError.

fixes https://github.com/trilogy-libraries/activerecord-trilogy-adapter/issues/52

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -96,6 +96,8 @@ static void trilogy_syserr_fail_str(int e, VALUE msg)
         rb_raise(Trilogy_ConnectionRefusedError, "%" PRIsVALUE, msg);
     } else if (e == ECONNRESET) {
         rb_raise(Trilogy_ConnectionResetError, "%" PRIsVALUE, msg);
+    } else if (e == EADDRNOTAVAIL) {
+        rb_raise(Trilogy_BaseConnectionError, "%" PRIsVALUE ": TRILOGY_DNS_ERROR", msg);
     } else {
         VALUE exc = rb_funcall(Trilogy_SyscallError, id_from_errno, 2, INT2NUM(e), msg);
         rb_exc_raise(exc);


jhawthorn:
I'm not sure I agree with the motivation here. A Trilogy::SyscallError is what will be raised here and what we'll raise for a number of other connection-related errors. It seems best to use the most specific error we can for the error we've encountered.

dhruvCW:
>I'm not sure I agree with the motivation here. A Trilogy::SyscallError is what will be raised here and what we'll raise for a number of other connection-related errors. It seems best to use the most specific error we can for the error we've encountered.

don't you think it would make sense to raise an exception that is derived from a `Trilogy::ConnectionError` base ? I am thinking in terms of a use case like active_record adapter where it might be better to handle all connection errors rather than specific ones ?

In this case though I am not particularly invested anymore since it arises from an invalid system setup and thus is more of an outlier than other errors.

jhawthorn:
I think it could make sense as a `Trilogy::ConnectionError`, but only if that was also done to the other `Trilogy::SyscallError::*` classes, which are effectively all "connection-level" errors.

thoughts @adrianna-chang-shopify @casperisfine ?

casperisfine:
> thoughts

Yeah, that's why I don't like the syscall errors, I'd like to be able to `rescue Trilogy::BaseConnectionError` or something, and inheriting from `SystemCallError` prevents that :/

adrianna-chang-shopify:
> Yeah, that's why I don't like the syscall errors, I'd like to be able to rescue Trilogy::BaseConnectionError

Yeah, I guess the idea with our approach was that consumers could generically rescue syscall errors as `Trilogy::Error`, but also use the raw syscall errs (e.g. `ECONNRESET`) for finer-tuned handling. Changing the syscall errors to include `Trilogy::BaseConnectionError` makes sense to me 👍

PR #100, opened 2023-07-11
Cast port to intrger in connection options

gmcgibbon:
Support string ports to make ENV config passing easier. Other gems like mysql2 support this use-case (https://github.com/brianmario/mysql2/blob/master/lib/mysql2/client.rb#L92), so I think trilogy should too.

### DIFF
--- a/contrib/ruby/lib/trilogy.rb
+++ b/contrib/ruby/lib/trilogy.rb
@@ -8,6 +8,7 @@ require "trilogy/encoding"
 
 class Trilogy
   def initialize(options = {})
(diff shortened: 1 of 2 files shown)


adrianna-chang-shopify:
There's a Ruby test that asserts that establishing a connection with `port` passed as a String raises a `TypeError`. I think we ought to remove it, but let's check with the GitHub folks to ensure they don't have a use-case for this.

gmcgibbon:
Ah, yeah good catch, removed!

adrianna-chang-shopify (approved):

Issue #72, opened 2023-05-04
Add high-level docs for the Ruby gem


composerinteralia:
https://github.com/github/trilogy/issues/71 was a good reminder that the gem does a whole bunch of things that currently have little or no documentation. With Trilogy soon to be an officially supported option for Rails, now is a good time to add some better documentation (expanding what we have in https://github.com/github/trilogy/blob/7ceee303b7d440f0b26cba65d00032bfb897c0ef/contrib/ruby/README.md would be a good place to start).

bensheldon:
Some things that we likely want to document, either through example or reasonable reference (i.e. anything that is Trilogy-specific, or database options/flags that are translated into Ruby-style; we shouldn't duplicately list every database options/flag that are documented elsewhere):

- `Trilogy.new` arguments
- `Trilogy` instance methods (the practical example already in the readme is a good starting place)
- Usage of `Trilogy::Result` 
- Error classes and their general meaning

composerinteralia:
It'd be nice to have supported versions of mysql and Ruby listed too (e.g. https://github.com/trilogy-libraries/trilogy/issues/225)

PR #104, opened 2023-07-18
Expand on Ruby docs for multi-statement queries

adrianna-chang-shopify:
For: https://github.com/trilogy-libraries/trilogy/issues/72

Pulls the multi-statement example into a sub-section under `Usage` and expands a bit more on the API. I'm starting work on the Ruby bindings for prepared statements, and while working on the docs I felt like we could elaborate a bit more on some of the APIs beyond just a code example.


composerinteralia (approved):
Thank you!

Issue #103, opened 2023-07-17
Use memset_s() instead of memset() to clear passwords


ArisBee:
I see [here](https://github.com/trilogy-libraries/trilogy/blob/640f990f26c12564935af16a96d9a9bf294d0a1d/src/client.c#L420) that the `memset()` function is used to clear passwords.

As passwords are sensitive information I would suggest to replace it with `memset_s()` so that they can't be recovered from memory.

More information on StakeOverflow: https://stackoverflow.com/questions/56565539/memset-s-what-does-the-standard-mean-with-this-piece-of-text

PR #106, opened 2023-07-19
Use memset_s to clear auth password.

dhruvCW:
There is a possibility that memset is optimised away, which would be problematic when clearing the password. Thus we replace memset with memset_s which has a far stricter language preventing this optimisation.

fixes https://github.com/trilogy-libraries/trilogy/issues/103

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -402,7 +402,11 @@ int trilogy_auth_switch_send(trilogy_conn_t *conn, const trilogy_handshake_t *ha
 void trilogy_auth_clear_password(trilogy_conn_t *conn)
 {
     if (conn->socket->opts.password) {
+        # ifdef __STDC_WANT_LIB_EXT1__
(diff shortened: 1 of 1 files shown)


src/client.c
```diff
@@ -417,7 +418,7 @@ int trilogy_auth_switch_send(trilogy_conn_t *conn, const trilogy_handshake_t *ha
 void trilogy_auth_clear_password(trilogy_conn_t *conn)
 {
     if (conn->socket->opts.password) {
-        memset(conn->socket->opts.password, 0, conn->socket->opts.password_len);
+        memset_s(conn->socket->opts.password, conn->socket->opts.password_len, 0, conn->socket->opts.password_len);
```
dbussink, line 421: This function is not always defined, so it would need a detection mechanism to see if it's available.

dhruvCW: yup realized too late that `memset_s` is not in posix. I believe `explicit_bzero` would be the alternative. Need to do an if def to handle this 😑

PR #110, opened 2023-08-01
Improve wait callback and timeout handling

jhawthorn:
This makes a couple improvements to how we wait for the socket to become ready for I/O.

First this refactors `TRILOGY_RB_TIMEOUT` to a "real" error status in the C library (though currently nothing in the C library raises it), which allows it to be negative like all the other errors (and avoids being confused with a successful read/write of size 1).

Next this makes `_cb_ruby_wait` return this new status code instead of SYSERR when there is a timeout, allowing it to propogate through `trilogy_sock_upgrade_ssl` and similar correctly. This allows differentiating between actually syserrors which set errno and timeouts. Also this can show the difference between `rb_wait_for_single_fd` returning `-1` (a syscall error, which I think in practice will be extremely rare) and 0 (a timeout).

Finally the last two commits ensure that the socket is shut down on either a socket timeout we see, or from an external exception (like `Timeout.timeout`). For the latter we must wrap the waiting in an `rb_protect`, which previously we were doing correctly for queries, but not for other operations (ex. `ping`, `change_db`). We have to shut down the socket because we've interrupted normal control flow and are likely either in the middle of a write (we've partly written our packet) or a read (there will be data sent from the server that we need to handle) so any further operations are invalid.

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -113,6 +113,9 @@ static void handle_trilogy_error(struct trilogy_ctx *ctx, int rc, const char *ms
         trilogy_syserr_fail_str(errno, rbmsg);
 
     case TRILOGY_TIMEOUT:
+        if (ctx->conn.socket != NULL) {
+            trilogy_sock_shutdown(ctx->conn.socket);
+        }
         rb_raise(Trilogy_TimeoutError, "%" PRIsVALUE, rbmsg);
 
     case TRILOGY_ERR: {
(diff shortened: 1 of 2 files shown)


contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -219,11 +239,25 @@ static int _cb_ruby_wait(trilogy_sock_t *sock, trilogy_wait_t wait)
...
+    int state = 0;
+    rb_protect(rb_trilogy_wait_protected, (VALUE)&args, &state);
+    if (state) {
+        trilogy_sock_shutdown(sock);
+        rb_jump_tag(state);
+    }
+
+    if (args.rc < 0)
         return TRILOGY_SYSERR;
+    if (args.rc == 0)
```
composerinteralia, line 257: I got confused for a second because in most other places `rc == 0` means `TRILOGY_OK`. I understand now that this is the return value of `rb_wait_for_single_fd`, where 0 means a timeout: https://github.com/ruby/ruby/blob/1642e0c39220e95ddb16b4cbbbe78f24507dfd48/include/ruby/io.h#L902. Possibly worth an inline comment in addition to what you have in the commit message?

composerinteralia (approved):

PR #112, opened 2023-08-09
Fixes to OpenSSL error handling

jhawthorn:
This fixes a number of issues with our OpenSSL integration.

First this ensures that the socket is shutdown (or closed post-#111) after error from OpenSSL. Previously we only closed the connection on `TRILOGY_OPENSSL_ERR` but not on a `TRILOGY_CLOSED_CONNECTION` or `TRILOGY_SYSERR` which could be returned/translated from an error in OpenSSL. As SSL holds its own buffers we must not attempt anything (including `SSL_shutdown`) after a failure from `SSL_read` or `SSL_write`.

To address that this also makes ALL exceptions (other than `TRILOGY_ERR` and `TRILOGY_MAX_PACKET_EXCEEDED`) shutdown the connection. These exceptions don't generally indicate something recoverable (if we get an error from a syscall other than EAGAIN, which we handle specially, we're likely to keep seeing it) and even if it was recoverable these exceptions happen in while we're in an inconsistent state (writing, or while expecting more data from the server), attempts to send a new packet from the start will be invalid.

Next this moves some translation based on `errno` from `client.c` to make it specific to raw sockets. I believe an OpenSSL socket will return `SSL_ERROR_WANT_WRITE` or `SSL_ERROR_WANT_READ` in any case the underlying socket returns EAGAIN, but that's not an assumption we should be making on the return value. OpenSSL has its own buffers and we can only retry when it tells us to. This also moves the translation of EPIPE to the raw socket, but also special cases it in `trilogy_syserr_fail_str` for backwards compatibility. `client.c` should be threating the socket like a black box not be assuming that it is backed by system calls.

Next this fixes checks on the return value from `SSL_read` and `SSL_write`. A return value of 0 is an error. Previously we incorrectly assumed that return value held the same meaning as from `read` or `write`, which is incorrect. Usually this happened on an unexpected EOF from the socket (which we'd happen to treat correctly, assuming the same meaning as `read`), but looking through the OpenSSL source 0 can mean a number of other errors. Instead this now adds a check for the case described in the [SSL_get_error's man page "BUG" section](https://www.openssl.org/docs/man1.1.1/man3/SSL_get_error.html).

> The SSL_ERROR_SYSCALL with errno value of 0 indicates unexpected EOF from the peer. This will be properly reported as SSL_ERROR_SSL with reason code SSL_R_UNEXPECTED_EOF_WHILE_READING in the OpenSSL 3.0 release because it is truly a TLS protocol error to terminate the connection without a SSL_shutdown().
> 
> The issue is kept unfixed in OpenSSL 1.1.1 releases because many applications which choose to ignore this protocol error depend on the existing way of reporting the error.

Additionally this clears any SSL errors inside `rb_trilogy_close`. We don't check whether `trilogy_close_send`/`trilogy_close_recv` return an error, which is (probably) fine: we're closing the socket, but we must clear any OpenSSL errors which are left on the stack from that read/write operation.

Finally, this adds a call to `ERR_clear_error()` before every OpenSSL operation. This is a workaround due to other libraries which incorrectly leave unhandled errors in the global (thread-local) OpenSSL error queue. This is a workaround other libraries also seem to have adopted (ones I know of: libpq and libcurl, the later of which ironically is responsible for causing this problem).

I would be happy to not look at OpenSSL for a while after this 😅

cc @adrianna-chang-shopify @casperisfine @composerinteralia @matthewd

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -94,6 +94,9 @@ static void trilogy_syserr_fail_str(int e, VALUE msg)
         rb_raise(Trilogy_ConnectionRefusedError, "%" PRIsVALUE, msg);
     } else if (e == ECONNRESET) {
         rb_raise(Trilogy_ConnectionResetError, "%" PRIsVALUE, msg);
+    } else if (e == EPIPE) {
+        // Backwards compatibility: This error class makes no sense, but matches legacy behavior
+        rb_raise(Trilogy_QueryError, "%" PRIsVALUE ": TRILOGY_CLOSED_CONNECTION", msg);
     } else {
         VALUE exc = rb_funcall(Trilogy_SyscallError, id_from_errno, 2, INT2NUM(e), msg);
         rb_exc_raise(exc);
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -994,6 +994,10 @@ static VALUE rb_trilogy_close(VALUE self)
         }
     }
 
+    // We aren't checking or raising errors here (we need close to always close the socket and free the connection), so
+    // we must clear any SSL errors left in the queue from a read/write.
+    ERR_clear_error();
+
     trilogy_free(&ctx->conn);
 
     return Qnil;
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -100,6 +100,17 @@ static void trilogy_syserr_fail_str(int e, VALUE msg)
     }
 }
 
+static int trilogy_error_recoverable_p(int rc)
+{
+    // TRILOGY_OPENSSL_ERR and TRILOGY_SYSERR (which can result from an SSL error) must shut down the socket, as further
+    // SSL calls would be invalid.
+    // TRILOGY_ERR, which represents an error message sent to us from the server, is recoverable.
+    // TRILOGY_MAX_PACKET_EXCEEDED is also recoverable as we do not send data when it occurs.
+    // For other exceptions we will also close the socket to prevent further use, as the connection is probably in an
+    // invalid state.
+    return rc == TRILOGY_ERR || rc == TRILOGY_MAX_PACKET_EXCEEDED;
+}
+
 NORETURN(static void handle_trilogy_error(struct trilogy_ctx *, int, const char *, ...));
 static void handle_trilogy_error(struct trilogy_ctx *ctx, int rc, const char *msg, ...)
 {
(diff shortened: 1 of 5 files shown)


contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -94,12 +94,26 @@ static void trilogy_syserr_fail_str(int e, VALUE msg)
         rb_raise(Trilogy_ConnectionRefusedError, "%" PRIsVALUE, msg);
     } else if (e == ECONNRESET) {
         rb_raise(Trilogy_ConnectionResetError, "%" PRIsVALUE, msg);
+    } else if (e == EPIPE) {
+        // Backwards compatibility: This error class makes no sense, but matches legacy behavior
+        rb_raise(Trilogy_QueryError, "%" PRIsVALUE ": TRILOGY_CLOSED_CONNECTION", msg);
```
jhawthorn, line 99: Just noting that I didn't want to change this exception in this PR, this is the exact exception that would have previously been raised: QueryError is our fallback and is what was raised by a `TRILOGY_CLOSED_CONNECTION`, which is what EPIPE was previously translated into.

Let's discuss separately how we can improve exceptions as I'm sure there is discussion to be had.

jhawthorn:
Pushed one extra commit to allow `SSL Error: unexpected eof while reading` inside `assert_raises_connection_error`, which should only occur on OpenSSL 3+ .

I think that makes sense to allow as the most specific error we have, but could see an argument that we should detect `SSL_R_UNEXPECTED_EOF_WHILE_READING` and translate it to the same error we'd see from `TRILOGY_CLOSED_CONNECTION`. What do y'all think?

contrib/ruby/test/client_test.rb
```diff
@@ -833,6 +839,10 @@ def test_configured_max_packet_above_server
     end
 
     refute_match(/TRILOGY_MAX_PACKET_EXCEEDED/, exception.message)
+
+    exception = assert_raises_connection_error do
```
adrianna-chang-shopify, line 843: Tiniest nit -- let's avoid reassigning here?
```suggestion
    assert_raises_connection_error do
```

adrianna-chang-shopify (approved):
The intricacies of OpenSSL sockets, oof 😅 These changes look good to me 👍 

Re
> I think that makes sense to allow as the most specific error we have, but could see an argument that we should detect SSL_R_UNEXPECTED_EOF_WHILE_READING and translate it to the same error we'd see from TRILOGY_CLOSED_CONNECTION. What do y'all think?

I'm in favour of raising this as a `Trilogy::SSLError` over a more generic connection closed error, given that OpenSSL is opting to treat it as an SSL protocol error. Are there concerns about the error class changing from OpenSSL 1.1.1 to OpenSSL 3?

jhawthorn:
> Are there concerns about the error class changing from OpenSSL 1.1.1 to OpenSSL 3?

I'm not especially concerned, but we should probably update Rails to consider it a retryable error.

PR #114, opened 2023-08-11
Ensure non-null socket before attempting to recv

ianks:
This PR adds some checks to ensure the socket ptr is non-null before attempting to close connection. Previously, there was an edge case which could happen when a connection was closed when during the `read_query_response` loop.

Although user's _reallllly_ shouldn't share a connection across multiple threads, at least with this patch they'll get a Ruby error rather than a segfault.

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -704,6 +704,10 @@ static VALUE read_query_response(VALUE vargs)
     int rc;
 
     while (1) {
+        if (ctx->conn.socket == NULL) {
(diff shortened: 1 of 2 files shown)


jhawthorn:
Thanks for the PR and especially the test. I agree both that a connection should not be used cross-tread but also that we should try to handle.

I'll poke at this next week. There are other spots we release the GVL and I'd like to check that those are safe, but also I think we should consider preventing against all attempts at invalid concurrent communication.

ianks:
> but also I think we should consider preventing against all attempts at invalid concurrent communication.

Couldn't agree more. I do still think this patch is worthwhile in it's own right - we should always test the socket field given it's nullable. Thanks for taking a look.

PR #117, opened 2023-08-22
Use expand_heap in GC.verify_compaction_references

javierhonduco:
As `double_heap` was deprecated in https://github.com/ruby/ruby/commit/a6dd859affc42b667279e513bb94fb75cfb133c1


jhawthorn:
Thanks for the change! I've seen this warning but didn't get around to fixing it 😅. I think we're still testing some older Rubies, could you make this change conditional on having a Ruby version which supports `expand_heap`?

javierhonduco:
Oops didn't realise that CI was testing for more versions

Issue #113, opened 2023-08-09
Improved Ruby error classes


jhawthorn:
cc @adrianna-chang-shopify @casperisfine @composerinteralia @dhruvCW @matthewd 

We made some improvements to the Ruby extension's error classes recently but based on recent discussions I think still have a way to go. I wanted to gather opinions in this issue and am happy to implement what we can come to a consensus on.

Problems I see:

* QueryError is our default fallback error class. I don't think this makes any sense, as this is a ProtocolError subclass but all errors other than TRILOGY_ERR are IMO connection-level errors. I think we should make `BaseConnectionError` the default error class for things which haven't been given a specific meaning (alternatively we could exhaustively define error classes for all trilogy status codes).
* "TRILOGY_CLOSED_CONNECTION". This error represents essentially an EOF over our socket, and it's one of the more common errors we come across. It's also the error we'll see if we attempt communication after a `trilogy_sock_shutdown`. Currently this appears as `QueryError: TRILOGY_CLOSED_CONNECTION` and we're forced to do string matching to detect it. I think this deserves its own error class, but also it should probably be distinct from `ConnectionClosed`, which is only raised from an explicitly closed connection. How about `Trilogy::EOFError`? For backwards compatibility I'd probably include "TRILOGY_CLOSED_CONNECTION" in the string.
  * Another possibility I'd like to entertain is deleting the translation of EPIPE errors to `TRILOGY_CLOSED_CONNECTION`, which would break compatibility with those doing string matching but I think otherwise is just better and more accurate information.
* SyscallError vs ConnectionError. As raised in https://github.com/trilogy-libraries/trilogy/pull/98, there's some confusion between SyscallError and BaseConnectionError. One option is that we can make `Trilogy::SyscallError` inherit from `ConnectionError` and then have only one thing to rescue, or we could better document that users should rescue both classes (I don't feel very strongly either way).
* I think `ProtocolError`'s name is confusing. I read this as "protocol violation" rather than "we were told over the protocol that an error occurred", and I've seen confusion from others about the same. I preferred the previous `DatabaseError` naming but am open to other options. How about `ServerError` "an error reported by the server"? Whatever we choose we can make the classes aliases of each other, so there should be minimal compatibility problems.

casperisfine:
I think the key thing to consider when designing error hierarchies, is the use case. 

Which errors do users need to rescue?

For network libraries my answer is generally that errors fall in two main groups:

[...]

PR #118, opened 2023-08-23
Introduce EOFError and make SyscallError < ConnectionError

jhawthorn:
This makes two of the changes there seems to be consensus around from https://github.com/trilogy-libraries/trilogy/issues/113

First this introduces `EOFError`, which represents `TRILOGY_CLOSED_CONNECTION`. It's a subclass of `BaseConnectionError`, but distinct from `ConnectionClosed` (which represents an explicit `.close` from the client's side).

Second this makes `Trilogy::SyscallError::*` all include the `ConnectionError` module, allowing `rescue ConnectionError` to catch most types of network communication issue.

### DIFF
--- a/contrib/ruby/lib/trilogy/error.rb
+++ b/contrib/ruby/lib/trilogy/error.rb
@@ -20,7 +20,7 @@ class Trilogy
       .select { |c| c.is_a?(Class) && c < SystemCallError }
       .each do |c|
         errno_name = c.to_s.split('::').last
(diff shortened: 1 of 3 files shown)


contrib/ruby/lib/trilogy/error.rb
```diff
@@ -112,7 +112,13 @@ class SSLError < BaseError
     include ConnectionError
   end
 
+  # Raised on attempt to use connection which was explicitly closed by the user
```
composerinteralia, line 115: Thank you! I've been confused by this in the past.

composerinteralia (approved):

PR #122, opened 2023-09-13
Set TCP_NODELAY

shanth96:
This PR sets the `TCP_NODELAY` flag on sockets.

Sometimes, when clients use SSL, Nagle's could cause increased connection latency. We were seeing around 40ms additional latency.

### DIFF
--- a/src/socket.c
+++ b/src/socket.c
@@ -195,6 +195,14 @@ static int raw_connect_internal(struct trilogy_sock *sock, const struct addrinfo
         return TRILOGY_SYSERR;
     }
 
+#ifdef TCP_NODELAY
+    if (sock->addr->ai_family != PF_UNIX) {
+        int flags = 1;
+        if (setsockopt(sock->fd, IPPROTO_TCP, TCP_NODELAY, (void *)&flags, sizeof(flags)) < 0) {
+            goto fail;
+        }
+    }
+#endif
     if (sock->base.opts.keepalive_enabled) {
         int flags = 1;
         if (setsockopt(sock->fd, SOL_SOCKET, SO_KEEPALIVE, (void *)&flags, sizeof(flags)) < 0) {


src/socket.c
```diff
@@ -195,6 +195,14 @@ static int raw_connect_internal(struct trilogy_sock *sock, const struct addrinfo
         return TRILOGY_SYSERR;
     }
 
+#ifdef TCP_NODELAY
+    if (sock->base.opts.keepalive_enabled) {
```
adrianna-chang-shopify, line 199: Should this be the nodelay option?
```suggestion
    if (sock->base.opts. tcp_nodelay_enabled) {
```

src/socket.c
```diff
@@ -195,6 +195,14 @@ static int raw_connect_internal(struct trilogy_sock *sock, const struct addrinfo
         return TRILOGY_SYSERR;
     }
 
+#ifdef TCP_NODELAY
```
adrianna-chang-shopify, line 198: Nit: Could we move this down with the other TCP options?

adrianna-chang-shopify (approved):

brendar:
Should we enable TCP_NODELAY by default? It seems like more of a footgun than a benefit for mysql connections and there is a solid precedent for it. libmysql appears to set it by default (based on strace), and golang sets it by default for all TCP connections.

@byroot do you have any opinions on this?

adrianna-chang-shopify:
Makes sense to me to set the flag by default -- although let's ensure that it can be disabled with `tcp_no_delay_enabled: false` so that applications can opt out if necessary.

cc @composerinteralia @jhawthorn @matthewd -- GitHub folks, any opinions here?

src/socket.c
```diff
@@ -195,6 +195,14 @@ static int raw_connect_internal(struct trilogy_sock *sock, const struct addrinfo
...
+    if (sock->base.opts.tcp_nodelay_enabled) {
+      int flags = 1;
+      if (setsockopt(sock->fd, IPPROTO_TCP, TCP_NODELAY, (void *)&flags, sizeof(flags)) < 0) {
+        goto fail;
+      }
+    }
+#endif
```
eileencodes, line 205: The indentation is off for C code, should be indented one more time.

```suggestion
#ifdef TCP_NODELAY
    if (sock->base.opts.tcp_nodelay_enabled) {
        int flags = 1;
        if (setsockopt(sock->fd, IPPROTO_TCP, TCP_NODELAY, (void *)&flags, sizeof(flags)) < 0) {
            goto fail;
        }
    }
#endif
```

byroot:
> @byroot do you have any opinions on this?

Yes. It's extremely common to set it by default on connections that are expected to be on LAN like a DB connection.

I do set it in `redis`, pretty sure `Dalli` sets it as well, etc.

shanth96:
> I think our code is also well suited to TCP_NODELAY, I believe we always send large constructed buffers (though that is worth double checking that we don't have consecutive writes which could have been batched).

In that case, it definitely makes sense to omit the option and set it by default - will make things a lot simpler too.

shanth96:
Updated

kovyrin:
Here is the effect from this change on our Vitess clients:

image

src/socket.c
```diff
@@ -195,6 +195,14 @@ static int raw_connect_internal(struct trilogy_sock *sock, const struct addrinfo
         return TRILOGY_SYSERR;
     }
 
+#ifdef TCP_NODELAY
+    if (sock->addr->ai_family != PF_UNIX) {
```
shanth96, line 199: Does this make sense?

jhawthorn: Looks good to me! We can't and don't need to apply this to unix sockets.

jhawthorn (approved):

PR #124, opened 2023-10-03
Add `Result#columns` to store field types

### LINKED ISSUE #99: No access to field types for a query result
In `mysql2`, a result has a `field_types` method to get access to the field types of the result. Trilogy doesn't seem to provide any way to get access to these in the Ruby layer. The C driver does provide it, but it's not exposed in Ruby. 

The Ruby driver does use it for casting purposes, so it's internally available. See also:

https://github.com/trilogy-libraries/trilogy/blob/640f990f26c12564935af16a96d9a9bf294d0a1d/contrib/ruby/ext/trilogy-ruby/cext.c#L738-L742

There the data is read into `column_info` but that data is ephemeral for the result object and not included in the final result.

We were looking to move to Trilogy, but we do need access to this field type information in some way or another. From purely that perspective, it doesn't have to be compatible 1:1 with how `mysql2` works, as long as it's available in some way.

I'm happy to contribute a PR to add this, but I don't really know what the preferred way of adding this would be. Should it be for example to `column_info` and then lazily create entries in a `field_types` method? Or should we build `field_types` directly like `fields` is built (although often it won't really be needed, as is clear since it's not present yet and already extensively used with Rails). 

Or should there be a totally different function than `field_types `that returns the types not as strings like `mysql2` does, but with integer constants + any of the additional flags / size attributes etc. for the type? If there's any preference with how to make this available, I'm also able to start a PR for that then.

ipc103:
Closes #99 

This PR adds a new `Result#columns` method which returns an array of `Result::Column` objects to represent column information such as name, length, and type. At some point this could replaced the `fields` method, but as of now was added as a non-breaking change. 

We added a new `Columns` class as well. When instantiating a new `Result` in the C extension, we assign a new instance of our `Columns` class which carries a pointer to a struct. After populating the properties of the result, we cache the `column_info` in a struct on the `Columns` Value and add the information in (including casting some integers into symbol values). We can then lazily-load the data out of the struct and populate the ruby array when calling `#columns`

Full disclosure: I've never written a C extension (or any C for that matter) before. Any and all feedback welcome - I added a few inline comments in the PR.

CC @composerinteralia @ywenc


contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -1095,6 +1147,45 @@ static VALUE rb_trilogy_write_timeout_set(VALUE self, VALUE write_timeout)
...
+#undef XX
+};
+
+const char *const trilogy_charset_names[] = {
+#define XX(name, code) [name] = (char *)#name + strlen("TRILOGY_CHARSET_"),
+    TRILOGY_CHARSETS(XX)
+#undef XX
+};
+
+static char *downcase(const char *str)
```
ipc103, line 1162: This is to convert the charset/type name into a lowercase symbol. I'm sure there is a better way to do this.

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -1095,6 +1147,45 @@ static VALUE rb_trilogy_write_timeout_set(VALUE self, VALUE write_timeout)
...
+static VALUE rb_trilogy_result_columns(VALUE self)
+{
+    struct trilogy_result_columns_ctx *trilogy_result_columns_ctx = get_trilogy_result_columns_ctx(self);
+    VALUE cols = rb_ary_new();
+    for (uint64_t i = 0; i < trilogy_result_columns_ctx->column_count; i++) {
+        VALUE obj = rb_funcall(
+            Trilogy_Result_Column, rb_intern("new"), 6, trilogy_result_columns_ctx->column_info[i].name,
+            rb_id2sym(rb_intern(downcase(trilogy_type_names[trilogy_result_columns_ctx->column_info[i].type]))),
+            rb_int_new(trilogy_result_columns_ctx->column_info[i].len), rb_int_new(trilogy_result_columns_ctx->column_info[i].flags),
+            rb_id2sym(rb_intern(downcase(trilogy_charset_names[trilogy_result_columns_ctx->column_info[i].charset]))),
```
ipc103, line 1182: Wrapping `rb_id2sym(rb_intern(...))` seemed to be _a_ way to get the actual string-like symbol value i.e. `:long`. There is probably a better way to do this but I couldn't find one 🤷

contrib/ruby/lib/trilogy/result.rb
```diff
@@ -28,6 +28,35 @@ def each(&bk)
...
 
+    def columns
+      @columns.all
+    end
+
     include Enumerable
+
+
+    class Columns
+      include Enumerable
```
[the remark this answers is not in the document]
ipc103: Good call - you are correct that we don't need this. Removed and updated the [previous commit](https://github.com/trilogy-libraries/trilogy/compare/401a44f01eb6fe7f03d2cc23cb844cb57b7111ef..d27a2c67647d2dfa4e7a4a8adb9191b94bb4331d)

(As an aside, not sure what the preference is for commit history here. Happy to squash these three down to a single commit if we prefer)

contrib/ruby/ext/trilogy-ruby/cext.c
dbussink: Does this leak memory? Afaik `rb_intern` doesn't take ownership of the allocated memory but ends up copying it?

ipc103: Thanks - given the fact that this already felt sub-optimal, I wonder if it makes more sense to just set an int here and then cast to the symbol on the Ruby class.

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -34,6 +34,11 @@ struct trilogy_ctx {
     VALUE encoding;
 };
 
+struct trilogy_result_columns_ctx {
+    struct column_info *column_info;
+    uint64_t column_count;
+};
```
byroot, line 40:
```suggestion
struct trilogy_result_columns_ctx {
    uint64_t column_count;
    struct column_info column_info[];
}; 
```

You could use a flexible array member to allocate the whole list at once and avoid pointer chasing and malloc churn.

PR #126, opened 2023-10-11
Fix contrib/readme.md

k0i:
# Motivation
delete the invalid `query_options` method that is deleted in #50 .


composerinteralia (approved):
Thanks!

PR #129, opened 2023-10-20
Trilogy is much slower than mysql2 to connect to an IP

casperisfine:
```
$ script/benchmark-connect
trilogy connect/close
                       107.000  i/100ms
mysql2 connect/close   761.000  i/100ms
trilogy connect/close
                          1.022k (±73.3%) i/s -    321.000  in  19.716812s
mysql2 connect/close      7.564k (±19.0%) i/s -     37.289k in   5.246187s

Comparison:
mysql2 connect/close:     7564.4 i/s
trilogy connect/close:     1022.4 i/s - 7.40x  slower
```

I'll try to investigate, but opening now FYI

### DIFF
--- a/contrib/ruby/script/benchmark-connect
+++ b/contrib/ruby/script/benchmark-connect
@@ -9,24 +9,29 @@ require "mysql2"
 
 DEFAULT_USER = ENV["MYSQL_USER"] || "root"
 DEFAULT_PASS = ENV["MYSQL_PASS"]
-DEFAULT_SOCK = ENV["MYSQL_SOCK"] || "/tmp/mysql.sock"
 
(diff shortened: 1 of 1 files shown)


casperisfine:
I profiled that benchmark, and somehow the difference seem to be mostly in `select`:

Capture d’écran 2023-10-20 à 12 27 03

Which is weird. Perhaps trilogy is failing to flush some packet or something?

matthewd:
> Now trilogy appear much faster than mysql2, but I still think something is weird

Huh.. yeah, while I might hope for faster, that sort of difference seems implausible. Unless perhaps it preemptively spins up OpenSSL or something? :confused:

PR #136, opened 2023-11-10
Fix Ruby version on CI

y-yagi:
Currently, Ruby version is specified unexist version(`2.8`). In this case, it seems that `ruby-build` use stable version(`3.2.2`). https://github.com/trilogy-libraries/trilogy/actions/runs/6646865109/job/18062157607#step:4:1205

But this pattern is already tested by other build. I assume this is just a typo of `2.7` that the last version of 2.x series.


y-yagi:
I misunderstood.  We need to fix Dockerfile first.

PR #139, opened 2023-12-11
Call `getaddrinfo` from Ruby

casperisfine:
Ref: https://bugs.ruby-lang.org/issues/19965

`getaddrinfo` is notorious for not having a timeout, which can make it hang for a very long time. Since Trilogy releases the GVL before calling it, that can cause the whole VM to be unresponsive and no longer respond to ctrl+c etc.

In 3.3.0, Ruby is now doing name resolution from a background thread, making `Socket.getaddrinfo` interruptible. So by doing the name resolution on the Ruby side, we should avoid such scenarios.

@jhawthorn @matthewd @composerinteralia @adrianna-chang-shopify what do you think?

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -25,7 +25,7 @@ static ID id_socket, id_host, id_port, id_username, id_password, id_found_rows, 
     id_ivar_affected_rows, id_ivar_fields, id_ivar_last_insert_id, id_ivar_rows, id_ivar_query_time, id_password,
     id_database, id_enable_cleartext_plugin, id_ssl_ca, id_ssl_capath, id_ssl_cert, id_ssl_cipher, id_ssl_crl, id_ssl_crlpath, id_ssl_key,
     id_ssl_mode, id_tls_ciphersuites, id_tls_min_version, id_tls_max_version, id_multi_statement, id_multi_result,
-    id_from_code, id_from_errno, id_connection_options, id_max_allowed_packet;
+    id_from_code, id_from_errno, id_connection_options, id_max_allowed_packet, id_ip_address;
 
 struct trilogy_ctx {
     trilogy_conn_t conn;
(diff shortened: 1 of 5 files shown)


contrib/ruby/test/client_test.rb
```diff
@@ -68,7 +68,9 @@ def test_trilogy_connection_options
       ssl_mode: 4,
       tls_min_version: 3,
     }
-    assert_equal expected_connection_options, client.connection_options
+    actual_options = client.connection_options.dup
+    actual_options.delete(:ip_address)
+    assert_equal expected_connection_options, actual_options
```
adrianna-chang-shopify, line 73: Can we assert `ip_address` on the expected connection options instead?

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -488,6 +488,11 @@ static VALUE rb_trilogy_connect(VALUE self, VALUE encoding, VALUE charset, VALUE
             Check_Type(val, T_FIXNUM);
             connopt.port = NUM2USHORT(val);
         }
+
+        if ((val = rb_hash_lookup(opts, ID2SYM(id_ip_address))) != Qnil) {
+            Check_Type(val, T_STRING);
+            connopt.ip_address = StringValueCStr(val);
```
adrianna-chang-shopify, line 494: Do we need to free `ip_address` as part of `_cb_raw_close`?

contrib/ruby/lib/trilogy.rb
```diff
@@ -8,10 +8,23 @@
...
     charset = Trilogy::Encoding.charset(mysql_encoding)
+
+    if options[:host]
+      addrs = begin
+        Socket.getaddrinfo(options[:host], options[:port], :PF_UNSPEC, :SOCK_STREAM)
+      rescue SocketError => error
+        raise Trilogy::BaseConnectionError, "Couldn't resolve host: #{options[:host].inspect}"
+      end
+
+      addrs.sort_by!(&:first) # Priority to IPv4
```
adrianna-chang-shopify, line 24: For my own understanding: does prioritizing IPv4 here match what happens when we were calling `getaddrinfo` directly in `trilogy_sock_resolve`? Were we resolving to a single address previously, or potentially to multiple? 🤔

adrianna-chang-shopify (commented):
This makes sense to me --  a couple comments / questions, but I'll let other folks weigh in.

contrib/ruby/ext/trilogy-ruby/cext.c
casperisfine: Good catch, I added it.

contrib/ruby/lib/trilogy.rb
casperisfine: Yeah, that's a good question to which I don't have a proper answer. This PR is more proof of concept territory because I'm not very good at networking, so there might be a much simpler / better way to do this.

The crux is really that I see processes stuck connecting to MySQL, and I highly suspect this is the cause.

jhawthorn:
I think it's a good idea to use Ruby as much as we can, but I'm also a bit unsure about the IP sorting

casperisfine:
> I'm also a bit unsure about the IP sorting

Yeah me too, I think I need to dig a bit more on how this works

PR #141, opened 2023-12-13
Update README.md with caching_sha2_password limitation

enortham:
Add caching_sha2_password limitation. I lost a lot of time trying to figure out why I couldn't authenticate and I think this should be in the README to save others in the future. Hopefully this will be fixed soon based on #26.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ It's currently in production use on github.com.
 
 * `trilogy_escape` assumes an ASCII-compatible connection encoding.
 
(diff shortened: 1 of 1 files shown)


andyjeffries:
I would change this to comment that it's the default configuration in the official Docker images too, otherwise people may not be aware of the setting and just think "I'm OK, I didn't change that option so I should be fine" and cause lots of wasted time.

enortham:
Thanks @andyjeffries. I added that it's the default for MySQL 8 which should affect not only Docker but anything that uses 8.

bensheldon:
Thank you for proposing this change! I made a few tweaks and will accept 👍🏻

andyjeffries:
Hi @bensheldon you removed the part about it being default for MySQL 8 and for the MySQL 8 docker build? That feels like key information.

andyjeffries:
Of course, this is something that @eileencodes is working on, so soon (for some definition...) this documentation change won't be necessary anyway.

Appreciate your work (all of you) on this great library, really looking forward to switching properly to it.

eileencodes:
Closing in favor of https://github.com/trilogy-libraries/trilogy/pull/165

PR #142, opened 2023-12-21
Return rc from trilogy_sock_wait_write in connect

jhawthorn:
In `raw_connect_internal` we call `trilogy_sock_wait_write`, which will error if we don't connect within the specified `write_timeout`. Previously, we ignored the error code from trilogy_sock_wait_write and always returned `TRILOGY_SYSERR`.

This only changes the error class and message returned and doesn't otherwise affect timeout behaviour.

**Before:**

```
>> Trilogy.new(host: "198.51.100.1", write_timeout: 1)
lib/trilogy.rb:18:in `_connect': Operation now in progress - trilogy_connect - unable to connect to 198.51.100.1:3306 (Trilogy::SyscallError::EINPROGRESS)
```

**After:**

```
>> Trilogy.new(host: "198.51.100.1", write_timeout: 0.1)
lib/trilogy.rb:18:in `_connect': Connection timed out - trilogy_connect - unable to connect to 198.51.100.1:3306 (Trilogy::TimeoutError)
```

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -577,9 +577,6 @@ static VALUE rb_trilogy_connect(VALUE self, VALUE encoding, VALUE charset, VALUE
     }
 
     int rc = try_connect(ctx, &handshake, &connopt);
-    if (rc == TRILOGY_TIMEOUT) {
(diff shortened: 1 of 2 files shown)


jhawthorn:
Just noticed that error message is less helpful 🤔 let me see if we can improve that too

composerinteralia (approved):

PR #143, opened 2023-12-21
Error class revamp (part 2)

jhawthorn:
@composerinteralia and I paired on making another pass over the errors. Partly inspired by complexities revealed in https://github.com/rails/rails/pull/50202

The most significant change is that this redefines `===` on our custom `SystemCallError` descendants to use the standard `Module#===` implementation. Without this they are hard to check for specifically because the Errno classes redefine `===` to make anything which responds to `errno` and returns the same one equal. As suggested by @byroot in https://github.com/rails/rails/pull/50202#discussion_r1408359465

This also removes the inheritance of `Trilogy::TimeoutError` from `Errno::ETIMEDOUT`, as that represents a timeout we have declared after waiting rather than an I/O timeout the kernel/stdlib is informing us about.

Lastly this deprecates `Trilogy::ConnectionRefusedError` and `Trilogy::ConnectionResetError`, replacing them with `Trilogy::SyscallError::ECONNREFUSED` and `Trilogy::SyscallError::ECONNRESET` (and making them aliases). This avoids having duplicate errors representing the same thing and makes all our "errno" errors share common implementation 😌. As a bonus this removes some C and replaces it with Ruby.

I think this makes some good progress in making errors easier to handle, but I do anticipate making at least a "part 3" 😅.

cc @adrianna-chang-shopify @matthewd

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -91,7 +91,7 @@ NORETURN(static void trilogy_syserr_fail_str(int, VALUE));
 static void trilogy_syserr_fail_str(int e, VALUE msg)
 {
     if (e == EPIPE) {
-        // Backwards compatibility: This error class makes no sense, but matches legacy behavior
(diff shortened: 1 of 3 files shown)


adrianna-chang-shopify (approved):
❤️

PR #152, opened 2024-01-24
Fix Docker build args for CI matrix; remove Ruby 1.9 tests; change base option to Ruby 3.3

### LINKED ISSUE #137: CI for old Rubies doesn't work correctly
Currently, it looks like trilogy tries to run against two old Ruby versions, `1.9` and `2.8`.

https://github.com/trilogy-libraries/trilogy/blob/e577853a1f40a681cd7fc91c3c4e97e69e31762c/.github/workflows/ci.yml#L24
https://github.com/trilogy-libraries/trilogy/blob/e577853a1f40a681cd7fc91c3c4e97e69e31762c/.github/workflows/ci.yml#L27

But, this is strange because Ruby `2.8` doesn't exist, the last version of 2.x series is `2.7`. It seems that the build for `1.9` and `2.8` uses Ruby 3.2.

[Build (5.7, debian:buster, 1.9)](https://github.com/trilogy-libraries/trilogy/actions/runs/6647200610/job/18062162641#step:4:5359)

I think  this is because `FROM` is set  after `ARG RUBY_VERSION`.
https://github.com/trilogy-libraries/trilogy/blob/e577853a1f40a681cd7fc91c3c4e97e69e31762c/Dockerfile#L2-L3

`FROM` creates a new scope and  you need to define `ARG` after `FROM` if you want to use it inside the scope.
https://docs.docker.com/engine/reference/builder/#scope

So the current Dockerfile ignores `RUBY_VERSION` arg and always use Ruby 3.2.

I'm not sure trilogy's support policy for old Ruby versions. So I submitted an issue for just for your information.

bensheldon:
Fixes #137

I couldn't find a working formula for Ruby 1.9.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,14 +17,11 @@ jobs:
       matrix:
         mysql: ["5.7", "8"]
         distribution: ["debian:buster", "ubuntu:focal", "ubuntu:bionic"]
-        ruby: ["3.2"]
+        ruby: ["3.3"]
         include:
           - mysql: "5.7"
             distribution: "debian:buster"
-            ruby: "1.9"
-          - mysql: "5.7"
-            distribution: "debian:buster"
-            ruby: "2.8"
+            ruby: "2.7.8"
     steps:
     - uses: actions/checkout@v4
     - name: docker login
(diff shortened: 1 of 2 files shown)


composerinteralia (approved):
Looks like we'll need to update the required builds to match. I removed all the outdated ones so you can get a green checkmark.

PR #155, opened 2024-01-30
Print warning when unknown options are passed to Trilogy#initialize

bensheldon:
Alternative to #151. 

I'm not confident I have caught all of the possible configuration options.

### DIFF
--- a/contrib/ruby/lib/trilogy.rb
+++ b/contrib/ruby/lib/trilogy.rb
@@ -17,7 +17,7 @@ class Trilogy
 
   def initialize(options = {})
     invalid_keys = options.keys - VALID_OPTIONS
-    $stderr.puts "WARNING: Unknown Trilogy options: #{invalid_keys.keys.join(", ")}" unless invalid_keys.empty?
+    $stderr.puts "WARNING: Unknown Trilogy options: #{invalid_keys.join(", ")}" unless invalid_keys.empty?
 
     options[:port] = options[:port].to_i if options[:port]
     mysql_encoding = options[:encoding] || "utf8mb4"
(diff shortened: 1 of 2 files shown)


contrib/ruby/lib/trilogy.rb
```diff
@@ -9,16 +9,15 @@
...
   ].freeze
 
   def initialize(options = {})
-    original_options = options.dup
-    options = VALID_OPTIONS.each_with_object({}) do |key, hash|
-      hash[key] = original_options.delete(key) if original_options.key?(key)
-    end
-    $stderr.puts "WARNING: Unknown Trilogy options: #{original_options.keys.join(", ")}" unless original_options.empty?
+    invalid_keys = options.keys - VALID_OPTIONS
+    $stderr.puts "WARNING: Unknown Trilogy options: #{invalid_keys.keys.join(", ")}" unless original_options.empty?
```
composerinteralia: I think Rails might pass through the whole db config, which includes keys that wouldn't be valid here. We should check that before merging this.

contrib/ruby/lib/trilogy.rb
```diff
@@ -7,7 +7,18 @@
 require "trilogy/encoding"
 
 class Trilogy
+  VALID_OPTIONS = %i[
+    host port path database username password encoding
```
bensheldon, line 11: What is the distinction between `socket` and `path` options? 

I see this:

https://github.com/trilogy-libraries/trilogy/blob/362ee5881a0a92867895d5bdfeb83165316bddc4/contrib/ruby/script/benchmark#L15-L16

..but also this:

https://github.com/trilogy-libraries/trilogy/blob/362ee5881a0a92867895d5bdfeb83165316bddc4/contrib/ruby/test/test_helper.rb#L61

composerinteralia: I don't think `path` is a valid option for either trilogy or mysql. Maybe it used to be called `path`?

PR #159, opened 2024-02-05
Use connect_timeout for initial connection

jhawthorn:
After the initial non-blocking connection call, we wait for the socket to become writable as a signal that it has been successful (non-blocking connects otherwise return immediately).

Previously this "writable" check used the write timeout instead of the connect timeout, which may have been confusing for users. IMO this was a bug introduced at some point prior to public release (at one time the writable check didn't exist and the connect_timeout worked as one would expect).

This commit adds a new TRILOGY_WAIT_CONNECT, which uses connect_timeout to wait for the connection to become writable. If connect_timeout isn't set it will use write_timeout for backwards compatibility.

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -239,7 +239,18 @@ static int _cb_ruby_wait(trilogy_sock_t *sock, trilogy_wait_t wait)
         wait_flag = RB_WAITFD_OUT;
         break;
 
+    case TRILOGY_WAIT_CONNECT:
+        // wait for connection to be writable
+        timeout = &sock->opts.connect_timeout;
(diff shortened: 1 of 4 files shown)


composerinteralia (approved):

nirvdrum:
FYI, Dalli started using `connect_timeout` recently and it's [been problematic](https://github.com/petergoldstein/dalli/pull/989). That's not to say you shouldn't use `connect_timeout`, but I wanted to highlight that it could be problematic through Ruby 3.4 due to the presence of _resolv-replace_ in the standard library.

jhawthorn:
Thanks for the warning. I think we're not going to hit any similar problems as this is our own `:connect_timeout` option not `Socket`'s (though that's good to keep in mind for the future as I am interested in exploring using a Ruby socket)

PR #160, opened 2024-02-15
Update mysql2 to a version that can compile with MySQL 8.3.

nirvdrum:
_mysql2_ 0.5.6 was released with a fix for a [compliation issue with MySQL 8.3](https://github.com/brianmario/mysql2/issues/1346). While it can be worked around by installing older MySQL client libraries, bumping the gem would obviate the need for doing so.


jhawthorn (approved):
Thanks!

PR #161, opened 2024-02-18
Rename test for `Trilogy#ping`

genya0407:
I renamed `test_trilogy_ping_after_close_returns_false` to `test_trilogy_ping_after_close_raises` to more accurately reflect its content as discussed here: https://github.com/trilogy-libraries/trilogy/pull/145#issuecomment-1880418956.

### DIFF
--- a/contrib/ruby/test/client_test.rb
+++ b/contrib/ruby/test/client_test.rb
@@ -80,7 +80,7 @@ class ClientTest < TrilogyTest
     ensure_closed client
   end
 
-  def test_trilogy_ping_after_close_returns_raises
+  def test_trilogy_ping_after_close_raises
     client = new_tcp_client
     assert client.ping
     client.close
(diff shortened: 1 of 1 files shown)


contrib/ruby/test/client_test.rb
```diff
@@ -80,7 +80,7 @@ def test_trilogy_ping
     ensure_closed client
   end
 
-  def test_trilogy_ping_after_close_returns_false
+  def test_trilogy_ping_after_close_returns_raises
```
composerinteralia, line 83: Thanks for updating this. I think we should get rid of `returns` from the test name now

```suggestion
  def test_trilogy_ping_after_close_raises
```

genya0407: 🙈 Thanks!

composerinteralia (approved):

PR #165, opened 2024-03-12
Implement caching_sha2_password for mysql 8+

eileencodes:
This PR implements caching_sha2_password for mysql 8. Note that we have
chosen on purpose to only implement the path where TLS or a unix socket is used. We will
not be implementing the non-TLS/non-unix socket path.
    
Co-authored-by: John Hawthorn <john@hawthorn.email>
Co-authored-by: Aaron Patterson (tenderlove) <tenderlove@ruby-lang.org>

Fixes: https://github.com/trilogy-libraries/trilogy/issues/26

---

**For testing:**

Point trilogy at this branch in your Gemfile:

```ruby
gem "trilogy", github: "trilogy-libraries/trilogy", branch: "caching_sha2_password", glob: "contrib/ruby/*.gemspec"
```

Update your mysql auth plugin / policy to use `caching_sha2_password` over `mysql_native_password`

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -6,7 +6,7 @@ on:
     tags:
       - v*
   pull_request:
-    
+
 jobs:
   test:
     name: Test
@@ -36,12 +36,15 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Setup MySQL
+      env:
+        MYSQL_VERSION: ${{ matrix.mysql }}
       run: |
         brew install mysql@${{ matrix.mysql }}
         (unset CI; brew postinstall mysql@${{ matrix.mysql }})
         brew services start mysql@${{ matrix.mysql }}
         sleep 5
         $(brew --prefix mysql@${{ matrix.mysql }})/bin/mysql -uroot -e 'CREATE DATABASE test'
+        [[ "$MYSQL_VERSION" == "8.0" ]] && $(brew --prefix mysql@${{ matrix.mysql }})/bin/mysql -uroot < docker-entrypoint-initdb.d/caching_sha2_password_user.sql
         $(brew --prefix mysql@${{ matrix.mysql }})/bin/mysql -uroot < docker-entrypoint-initdb.d/native_password_user.sql
         $(brew --prefix mysql@${{ matrix.mysql }})/bin/mysql -uroot < docker-entrypoint-initdb.d/x509_user.sql
     - name: Install dependencies
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -156,6 +156,10 @@ static void handle_trilogy_error(struct trilogy_ctx *ctx, int rc, const char *ms
         rb_raise(Trilogy_EOFError, "%" PRIsVALUE ": TRILOGY_CLOSED_CONNECTION", rbmsg);
     }
 
+    case TRILOGY_UNSUPPORTED: {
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -2,10 +2,6 @@ name: CI
 
 on:
   push:
-    branches:
-      - main
-    tags:
-      - v*
```
jhawthorn, line 8: Oops. I didn't mean to include this change I was doing it for testing 😓

.github/workflows/macos.yml
```diff
@@ -36,12 +36,15 @@ jobs:
...
     - name: Setup MySQL
+      env:
+        MYSQL_VERSION: ${{ matrix.mysql }}
       run: |
         brew install mysql@${{ matrix.mysql }}
         (unset CI; brew postinstall mysql@${{ matrix.mysql }})
         brew services start mysql@${{ matrix.mysql }}
         sleep 5
         $(brew --prefix mysql@${{ matrix.mysql }})/bin/mysql -uroot -e 'CREATE DATABASE test'
+        [[ "$MYSQL_VERSION" == "8.0" ]] && $(brew --prefix mysql@${{ matrix.mysql }})/bin/mysql -uroot < docker-entrypoint-initdb.d/caching_sha2_password_user.sql
```
dbussink, line 47: Should this be any `8.x` version and later? Since we also have versions up to 8.3.0 today for MySQL and each quarter there's a new one.

eileencodes: This is the image name and we don't have a matrix for `5.7`, `8.0`, and `8.3`. If we added 8.3 and then this wasn't updated, the build would fail. I'm not against doing a `>=` but it's also not broken currently. 🤷🏼‍♀️

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -156,6 +156,10 @@ static void handle_trilogy_error(struct trilogy_ctx *ctx, int rc, const char *ms
         rb_raise(Trilogy_EOFError, "%" PRIsVALUE ": TRILOGY_CLOSED_CONNECTION", rbmsg);
     }
 
+    case TRILOGY_UNSUPPORTED: {
+        rb_raise(Trilogy_QueryError, "%" PRIsVALUE ": TRILOGY_UNSUPPORTED", rbmsg);
```
composerinteralia, line 160: QueryError doesn't seem quite right to me. Maybe:

```suggestion
        rb_raise(Trilogy_BaseConnectionError, "%" PRIsVALUE ": TRILOGY_UNSUPPORTED", rbmsg);
```

contrib/ruby/test/client_test.rb
```diff
@@ -23,13 +23,86 @@ def test_trilogy_connect_tcp_string_host
...
+    options = {
+      host: DEFAULT_HOST,
+      port: DEFAULT_PORT,
+      username: "caching_sha2",
+      password: "password",
+      ssl: false,
+      ssl_mode: 0
+    }
+
+    err = assert_raises Trilogy::QueryError do
```
composerinteralia, line 83:
```suggestion
    err = assert_raises Trilogy::ConnectionError do
```

composerinteralia (approved):
One thought about the "unsupported" error class, but otherwise makes sense to me.

contrib/ruby/ext/trilogy-ruby/cext.c
eileencodes: I don't recall why I picked QueryError 🤔 I think I saw it in the output and assumed it was not able to be changed. Anyway, I've fixed this so I think the PR is good to go.

jhawthorn (approved):

Issue #157, opened 2024-02-01
Support for cleartext password


danschultzer:
It looks like trilogy doesn't allow for[ cleartext password](https://dev.mysql.com/doc/refman/8.0/en/cleartext-pluggable-authentication.html) and will always hash the password:

https://github.com/trilogy-libraries/trilogy/blob/362ee5881a0a92867895d5bdfeb83165316bddc4/src/protocol.c#L569-L577

We're using RDS IAM auth and we need to be able to send the token as a cleartext password. Is this something trilogy should support or is there a good reason for not being able to send cleartext password?

PR #171, opened 2024-04-03
Implement support for cleartext authentication plugin

wildmaples:
Resolves: https://github.com/trilogy-libraries/trilogy/issues/157
Redo of: https://github.com/trilogy-libraries/trilogy/pull/168

The server doesn't request for cleartext plugin in the first handshake as it is not a default plugin. The client should only provide cleartext password when the server requests an auth switch to it, and the client has `enable_cleartext_plugin` option enabled. If the client does not have cleartext auth plugin enabled, we will raise a `Trilogy::AuthPluginError`. 

This is the authentication flow for a server that uses the cleartext client side plugin (https://github.com/trilogy-libraries/trilogy/pull/168#discussion_r1548811735): 
> Server sends handshake, with auth plugin caching_sha2_password
> Client sends handshake response, still with auth plugin caching_sha2_password
> Server sends auth switch request, now with auth plugin mysql_clear_password
> Client sends auth switch response with just the cleartext password and nothing else
> Server sends OK and we are connected

This PR introduces these changes: 
- Supports cleartext auth plugin! 
- Raises when the server auth switches to a plugin that Trilogy does not support. Note that at the moment Trilogy supports mysql_native_password, caching_sha2_password and now mysql_clear_password. This is the recommended behaviour according to mysql [docs "Authentication Method Mismatch"](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase.html#sect_protocol_connection_phase_auth_method_mismatch).
- Adds a test dependency on this [mysql test server plugin](https://github.com/mysql/mysql-server/blob/824e2b4064053f7daf17d7f3f84b7a3ed92e5fb4/plugin/auth/test_plugin.cc#L176-L217). It's unclear if the file will stick around in the long term but seeing that it is only a test dependency and makes our lives easier than the alternative testing options, we are inclined to just use it for the time being. 

cc: @ngan @composerinteralia


src/protocol.c
```diff
@@ -600,21 +600,32 @@ int trilogy_build_auth_packet(trilogy_builder_t *builder, const char *user, cons
...
+            if (enable_cleartext_plugin) {
+                CHECKED(trilogy_builder_write_buffer(builder, pass, pass_len));
+            } else {
+                return TRILOGY_AUTH_PLUGIN_ERROR;
+            }
         } else {
-            trilogy_pack_scramble_native_hash(scramble, pass, pass_len, auth_response, &auth_response_len);
+            if (!strcmp("caching_sha2_password", auth_plugin)) {
+                trilogy_pack_scramble_sha2_hash(scramble, pass, pass_len, auth_response, &auth_response_len);
+            } else if (!strcmp("caching_sha2_password", auth_plugin)) {
```
composerinteralia, line 619:
```suggestion
            } else if (!strcmp("mysql_native_password", auth_plugin)) {
```

ngan:
I created an issue on the MySQL docker repo to see if they'd be open to building and tagging a `mysql-community-test` variant that we can use. If they do, we can switch to that and get the test auth plugin for free. If they don't, we can keep what we have (checked-in plugins for various mysql version), or build our own MySQL image.
 
https://github.com/docker-library/mysql/issues/1040

composerinteralia (approved):
This looks great to me, but I'll leave it for a bit in case anybody else has feedback.

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -1202,6 +1210,9 @@ RUBY_FUNC_EXPORTED void Init_cext(void)
     Trilogy_EOFError = rb_const_get(Trilogy, rb_intern("EOFError"));
     rb_global_variable(&Trilogy_EOFError);
 
+    Trilogy_AuthPluginError = rb_const_get(Trilogy, rb_intern("AuthPluginError"));
+    rb_global_variable(&Trilogy_AuthPluginError);
```
casperisfine, line 1214:
```suggestion
    rb_global_variable(&Trilogy_AuthPluginError);
    Trilogy_AuthPluginError = rb_const_get(Trilogy, rb_intern("AuthPluginError"));
```

Global variables should be registered before they are assigned. The windows for a bug is extremely small, so most gems get away with it, but we should correct that.

test/auth_plugins/5.7/auth_test_plugin.so
casperisfine, line 1: This probably shouldn't be committed?

adrianna-chang-shopify: Should we be building our own MySQL image instead of committing the plugins?

adrianna-chang-shopify (commented):
Thanks for working on this! Would you mind squashing your commits?

My preference would be for us to build our own MySQL image extending `mysql-community-test` over checking in those plugins, but don't feel strongly if folks would prefer to ship as is.

composerinteralia:
> Would you mind squashing your commits?

I do like a clean git history, but we haven't historically been as opinionated about this as (for example) Rails.

test/auth_plugins/5.7/auth_test_plugin.so
ngan: I think we can and should, but that would be quite a big detour from the original intent of this PR. I suggest we land as-is then I can spend some time refactoring the test setup.

adrianna-chang-shopify: Follow up sounds good to me 👍 Thanks!

adrianna-chang-shopify (approved):
> I do like a clean git history, but we haven't historically been as opinionated about this as (for example) Rails.

I'll defer to you folks, I wasn't sure if we were following Rails' stance of squash-everything-before-merge! 🙇‍♀️

contrib/ruby/test/auth_test.rb
```diff
@@ -0,0 +1,16 @@
+require "test_helper"
+
+class AuthTest < TrilogyTest
```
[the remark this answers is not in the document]
ngan: Done!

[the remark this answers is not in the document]
ngan: Actually, I could only add it for native password. We don't use cache_sha2 for mysql 5 so the ruby test fails there. I'm going to revisit this after this PR.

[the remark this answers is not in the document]
composerinteralia: I think we already have that test elsewhere, but we can sort it out later. We might want to move all connection-related stuff into something like ConnectionTest.

composerinteralia:
Thank you!

test/auth_plugins/5.7/auth_test_plugin.so
ngan: This has been addressed here: https://github.com/trilogy-libraries/trilogy/pull/174

PR #172, opened 2024-04-04
Use "docker compose" for Ubuntu

ngan:
This should fix the `docker-compose: command not found` errors in build, at least for Ubuntu. I'm curious to see if debian supports `docker compose` (v2) well. We might have to branch logic based on dist.


composerinteralia (approved):
Thanks!

PR #175, opened 2024-04-06
Use mysql:8.0-debian image tag for CI

ngan:
The `mysql:8-debian` tag is deprecated and is no longer receiving updates. The recommendation is to use the more explicit `mysql:8.0-debian` tag. This was discovered during (and extracted from) my work in https://github.com/trilogy-libraries/trilogy/pull/174.

This is the reference to the discovery:
https://github.com/docker-library/mysql/issues/1040#issuecomment-2040590043

> [!IMPORTANT] 
> Landing this means that we'll also need to change the required checks for this repo since the name is now `8.0` instead of `8`.

### Breakdown of changes

This wasn't a simple "let's change `8` to `8.0`"...

The newer implementation of [mysql's entrypoint file](https://github.com/docker-library/mysql/blob/master/docker-entrypoint.sh) does this thing where it checks to see if the data directory already exists. If it does, then the setup does nothing.
https://github.com/docker-library/mysql/blob/db3fdfbab4caa033f02126f525cb6dfb5a3c1b0c/docker-entrypoint.sh#L380-L381

This means that we can no longer [mount a shared directory at `/var/lib/mysql`](https://github.com/trilogy-libraries/trilogy/blob/6892cd0648fedd673c626eae66e04bd1544373eb/docker-compose.yml#L15-L16) otherwise the entrypoint script will not initialize the database. The purpose of this shared mount between the `db` and the `app` services was so that the `app` has access to the certificate files that we generate within the `db` container.

My workaround for this was to mount the directory somewhere else (`/mysql-certs`) and then change our `generate_keys.sh` script to [copy the generated certificates](https://github.com/trilogy-libraries/trilogy/pull/175/files#diff-b4fc1e6483f60131b0b80bf32aeecd1541e55adbf3a033b43c2a4c872b1ddb21R57-R60) from `/var/lib/mysql` to `/mysql-certs`. Since `generate_keys.sh` is executed as the `mysql` user, instead of `root`, we have to do a docker trick where we ensure that the directory to be mounted is [pre-created and `chmod`-ed with `777` _before_ it's mounted](https://github.com/trilogy-libraries/trilogy/pull/175/files#diff-5cd6595e4f0514d82e0ecd2c56a4a045e17a3adbe21dc88af6d875ea0c4f3815R53-R58).


ngan:
@composerinteralia this is ready for review. The PR check status is stuck in pending since the names of the required checks has changed. You’ll need to update that the required checks before landing this PR, I think.

composerinteralia (approved):

PR #176, opened 2024-04-08
Point mysql to certs instead of copying them

ngan:
In https://github.com/trilogy-libraries/trilogy/pull/175, I made it so that the certs were copied from `/var/lib/mysql/*.pem` into `/mysql-certs/*`. This felt awkward at the time but I wanted to defer the fix since it wasn't that important. I'm revisit this now and in this PR, we're going to tell mysql where to find the certificates instead.

This means that we now generate the certificates directly into `/mysql-certs/*` -- so no more copying!

### DIFF
--- a/script/cibuild
+++ b/script/cibuild
@@ -35,6 +35,7 @@ output_fold() {
 function cleanup() {
   echo
   echo "::group::Shutting down services..."
+  docker compose logs db
   docker compose down --volumes
   echo "::endgroup::"
 }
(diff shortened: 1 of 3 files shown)


script/cibuild
```diff
@@ -35,6 +35,7 @@ output_fold() {
 function cleanup() {
   echo
   echo "::group::Shutting down services..."
+  docker compose logs db
```
ngan, line 38: I added this because it helps a lot with debugging. It's a few extra log lines so figured we should just keep it instead of having to turn it on/off constantly.

composerinteralia (approved):

PR #178, opened 2024-04-08
Compartmentalize mysql-related files

ngan:
Simple housekeeping change to move all mysql-related build files into `test/mysql/`. Improvements this gives us:
1. Cleans up the root directory by removing `docker-entrypoint-initdb.d` and `Dockerfile.mysql`.
2. Move MySQL configurations into a `.cnf` file that gets automatically loaded at MySQL boot (`conf.d/build.cnf`). This gives us a lot of space so that we can start [adding comments](https://github.com/trilogy-libraries/trilogy/pull/178/files#diff-641600f44e7991ef2dd0df1c964429bb65a3441dea08b08d57e1d316e0e63a63R12-R17) as to why we are setting various configurations. As we start adding more MySQL options to the `command`, `docker-compose.yml` becomes too cluttered.
3. We can now rename our MySQL Dockerfile to just plain ol' `Dockerfile` without sacrificing clarity.


bensheldon:
Thank you for doing this cleanup!

What would you think about moving this under the `/test` directory? I think that might make it clearer that this is build/test harness related and not something that's specifically functional to the library itself.

ngan:
@bensheldon I debated that myself! No objections from me. Done.

bensheldon (approved):

bensheldon:
I didn't see the `Dockerfile.mysql` referenced in any of the other local development scripts 👍🏻

ngan:
Yup. All for CI

PR #182, opened 2024-04-10
Clear password without memset

ngan:
Fixes https://github.com/trilogy-libraries/trilogy/issues/103
Closes https://github.com/trilogy-libraries/trilogy/pull/182

Instead of using `memset()` to clear the password, clear the password byte by byte in a loop. Additionally, we mark the `char*` as `volatile` to ensure that the compiler does not optimize away the memory access or the clearing operation.

### DIFF
--- a/script/cibuild
+++ b/script/cibuild
@@ -55,7 +55,7 @@ export DISTRIBUTION_SLUG
 # and chmod this directory on the host so that the permissions are persisted when the
 # the directory is mounted in the containers. Since the mysql container runs as a non-root
 # user, we need to ensure that the directory is writable by all users.
-mkdir tmp/mysql-certs
+mkdir -p tmp/mysql-certs
 chmod 777 tmp/mysql-certs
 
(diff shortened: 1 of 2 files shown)


script/cibuild
```diff
@@ -55,7 +55,7 @@ export DISTRIBUTION_SLUG
 # and chmod this directory on the host so that the permissions are persisted when the
 # the directory is mounted in the containers. Since the mysql container runs as a non-root
 # user, we need to ensure that the directory is writable by all users.
-mkdir tmp/mysql-certs
+mkdir -p tmp/mysql-certs
```
ngan, line 58: Making this change because it was erring when I run this script locally multiple times.

ngan:
I talked to @composerinteralia about this at RailsConf and we were unsure why the password is cleared at all. It's still available in Ruby, so why bother clearing it in C? We need the password to stick around since it's needed/used for reconnecting, etc.

PR #187, opened 2024-05-13
Add rbmsg back to TRILOGY_ERR

skipkayhil:
This was previously [removed][1] when error classes were refactored, but it doesn't seem to have been done so intentionally.

When debugging a previous [fix][2], this missing context would have helped narrow down the problem initially as the rbmsg includes where the error was originating.

[1]: https://github.com/trilogy-libraries/trilogy/commit/f9ee855543de22d9b2e5a3714439458ade2e1717
[2]: https://github.com/trilogy-libraries/trilogy/commit/cb788be650d2f85502f5728c5fb3b287eb1f9313

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -133,7 +133,8 @@ static void handle_trilogy_error(struct trilogy_ctx *ctx, int rc, const char *ms
         rb_raise(Trilogy_TimeoutError, "%" PRIsVALUE, rbmsg);
 
     case TRILOGY_ERR: {
-        VALUE message = rb_str_new(ctx->conn.error_message, ctx->conn.error_message_len);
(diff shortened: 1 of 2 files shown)


skipkayhil:
I'm also curious if these lines should be changed to `trilogy_auth_switch_recv`

https://github.com/trilogy-libraries/trilogy/blob/004d02a638ba8cd69933feaee8808b72a7c97451/contrib/ruby/ext/trilogy-ruby/cext.c#L376-L381

While there is no `trilogy_auth_switch_recv` method, making the `rbMsg` here different than the one in `authenticate` would also be helpful for tracking down the line raising an error.

adrianna-chang-shopify (approved):
Thanks, yeah don't think we intended to remove this! 😅 

> I'm also curious if these lines in auth_switch should be changed to trilogy_auth_switch_recv

I think we should still reference the call that failed, but maybe we could add more context to the message, like `trilogy_auth_recv on authenticate` vs `trilogy_auth_recv on auth_switch`?

PR #190, opened 2024-06-10
Add `last_result_size` to ruby connection

EvilGenius13:
# Add `last_result_size` to ruby bindings

This PR introduces a new method `last_result_size` to the Trilogy gem's Ruby bindings. This method allows users to retrieve the exact byte size of the last result received from a query. The byte size includes not only the raw data but also headers, metadata, and any other protocol overhead.

### Main changes:
    - Added a new method `rb_trilogy_last_result_size` to surface the `conn.recv_buff_len` as a Ruby method.

### Note on the test
The test checks that the `last_result_size` falls within a specific range (150 to 170 bytes). The reason for using a range instead of an exact value is to account for potential changes in the future. For example, changes in how headers or metadata are sent could affect the total byte size. The range provides a buffer to accommodate any of these changes.

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -1131,6 +1131,8 @@ static VALUE rb_trilogy_server_status(VALUE self) { return LONG2FIX(get_open_ctx
 
 static VALUE rb_trilogy_server_version(VALUE self) { return rb_str_new_cstr(get_open_ctx(self)->server_version); }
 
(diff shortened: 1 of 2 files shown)


jhawthorn:
Thanks for the PR. Could you say more about what the motivation is for this?

I think (though I'm not 100% sure) that looking at `recv_buff_len` will tell us the size of the last row rather than all rows (which is what I would consider the "result").

composerinteralia:
I think `recv_buff_len` is going to be however many bytes we last read, which could be multiple MySQL packets but could also be an incomplete packet.

DerekStride:
We're trying to instrument the amount of data received over the network for each query.

PR #194, opened 2024-07-18
Add the new VECTOR type

dbussink:
MySQL 9.0 added a new vector type at the protocol level, so this adds support for it. It is a binary length encoded buffer, so treat it as such.

### DIFF
--- a/src/protocol.c
+++ b/src/protocol.c
@@ -926,6 +926,8 @@ int trilogy_build_stmt_execute_packet(trilogy_builder_t *builder, uint32_t stmt_
             case TRILOGY_TYPE_VAR_STRING:
             case TRILOGY_TYPE_STRING:
             case TRILOGY_TYPE_GEOMETRY:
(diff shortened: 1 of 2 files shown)


src/protocol.c
```diff
@@ -926,6 +926,8 @@ int trilogy_build_stmt_execute_packet(trilogy_builder_t *builder, uint32_t stmt_
             case TRILOGY_TYPE_VAR_STRING:
             case TRILOGY_TYPE_STRING:
             case TRILOGY_TYPE_GEOMETRY:
+            case TRILOGY_TYPE_JSON:
```
dbussink, line 929: This was missing `JSON` for some reason (probably bad copy paste), so I added it here for completeness.

dbussink:
See also https://dev.mysql.com/doc/dev/mysql-server/latest/field__types_8h.html for the value here.

jhawthorn (approved):
Looks great. Thank you!

PR #196, opened 2024-07-30
Add Rails setup to README

TALlama:
Previously, there was no documentation for how to use the adapter in Rails. With a local MySQL install it's pretty easy - you swap the `adapter` and Trilogy will use the `socket` just like `mysql2` does. But if the app is in Docker, a simple `adapter` change isn't enough; you'll get an `caching_sha2_password requires either TCP with TLS or a unix socket` error that does not tell you how to set up TCP with TLS. You need to set `ssl_mode`, too.

This commit adds a new section to the `contrib/ruby/README.md` to make all this clear.

If there's any additional context that should be included here, I'd be happy to add it. If this change should be in a different file (or a different spot in this file), I can do that, too!

### DIFF
--- a/contrib/ruby/README.md
+++ b/contrib/ruby/README.md
@@ -22,6 +22,37 @@ Or install it yourself as:
 $ gem install trilogy
 ```
 
+## Rails Setup
+
+After installation, you need to make sure your `config/database.yml` file is using Trilogy.
+
+If you're running MySQL on the same box as Rails, you can use a socket file:
+
+```
+development:
+  adapter: trilogy
+  encoding: utf8mb4
+  database: blog_development
+  pool: 5
+  username: root
+  password:
+  socket: /tmp/mysql.sock
+```
+
+If your MySQL is in another container or across the network, specify the `host` and make sure to require TLS connections:
+
+```
+development:
+  adapter: trilogy
+  encoding: utf8mb4
+  database: blog_development
+  pool: 5
+  username: root
+  password:
+  host: db
+  ssl_mode: :required
+```
+
 ## Usage
 
 ``` ruby


bensheldon:
@TALlama I think this would best live in the Rails docs, because that is where the Active Record Trilogy Adapter (as of Rails 7.1) lives.

TALlama:
OK; I'll open a PR over there. But I'm now unsure this commit is correct; adding `ssl_mode` into the config seems to break `mysqldump` (it errors out with `mysqldump: unknown variable 'ssl-mode=required'`).

The Rails guides don't seem to have any documentation on how Trilogy should be used, save the [7.1 release notes](https://github.com/rails/rails/blob/v7.1.3.4/guides/source/7_1_release_notes.md#introduce-adapter-for-trilogy) which don't mention the TLS requirement and don't do anything to enable it (the snippet there will just lead you into the TLS error message). Are you aware of any correct examples, since I'm now doubting my ability to write one?

bensheldon:
yeah, it's not exactly clear. You can see how the ssl-modes are configured in the adapter:

https://github.com/rails/rails/blob/fb46bd0e71a46c59c512cb8282b226bd39eb07ee/activerecord/lib/active_record/connection_adapters/trilogy_adapter.rb#L37-L44

The mapping is like `"required" -> SSL_MODE_[REQUIRED] -> ::Trilogy::SSL_REQUIRED_NOVERIFY -> 4` (where the Trilogy numerical constant is [here](https://github.com/trilogy-libraries/trilogy/blob/e84501b427201c8a2cb68dc5111d0c3bd709cca0/inc/trilogy/socket.h#L27)).

> adding `ssl_mode` into the config seems to break `mysqldump`

I think then maybe the constant should be `REQUIRED` because `mysqldump --ssl-mode=REQUIRED` should work 😕

TALlama:
`rake db:schema:dump` doesn't like the all-caps version, either. Same error, but all-caps now.

image

I may have the MariaDB client package, which [doesn't support `--ssl-mode`](https://jira.mariadb.org/browse/MDEV-22129) and [is doing their own thing instead](https://jira.mariadb.org/browse/MDEV-28634). Given that the code you posted only looks at the `:ssl_mode` key to determine if it's a secure connection, I'm not sure there's a way to get Trilogy and the MariaDB client to play nicely with each other; I may have to swap out the client package (if I can confirm that's really what's going on).

TALlama:
Switching the client package away from MariaDB did the trick. I think this conversation is a good set of breadcrumbs should anyone else come looking, but I'll close this up now.

Issue #189, opened 2024-06-05
[Ruby] Add a Pipelining API?


casperisfine:
### Context

Since MySQL server is generally hosted on a different machine, an important performance factor is the number of roundtrips you have to perform.

e.g. if you have a `1ms` round trip time, and 3 queries to perform in a transaction, you'll have to do 5 full roundtrips for a total of 5ms. Sometimes there is no way around this because you need the result of the previous query to perform the next one, but sometimes you don't.

The classic technique for minimizing the cost of roundtrips, is to send multiple queries in a row without immediately waiting for the response, AKA pipelining

![image](https://github.com/trilogy-libraries/trilogy/assets/19192189/af25f35f-8037-480a-a337-e9277177e189)

### Existing Solution with Trilogy

Right now some people are essentially doing poor man's pipelining by enabling `MULTI_STATEMENTS` and concatenating their queries together. It works but is wonky and a bit unsafe, as `MULTI_STATEMENTS` is off by default for a good reason, it prevent various injection vulnerabilities.

#### Mysql2

The `mysql2` gem has some limited support for pipelining, which they refer to as async: https://github.com/brianmario/mysql2?tab=readme-ov-file#async

It's a bit wonky because they only offer to queue the query and then block on `async_result`, but I suppose it works.

#### PG

The `pg` gem has some support for a pipeline API since PostgreSQL 14 (2021), I haven't tried it, and it seems a bit wonky, but it's the same idea. You start a pipeline with `#enter_pipeline_mode`, queue some queries in that pipeline, and can flush the pipelined queries whenever you want so they are sent all at once.

Here's the spec for it in the gem: https://github.com/ged/ruby-pg/blob/d072b21852865ecb84e6345df11d68eed50702bb/spec/pg/connection_spec.rb#L2009-L2122

#### Other databases

While pipelining isn't very commonly used with RDBMS, it has been used extensively for years with `redis-rb` and `redis-client`. 

`redis-client` only offer a simple block based API, and return all the results at once in an Array:

```ruby
result_1, result_2 = client.pipelined do |pipeline|
  pipeline.call("COMMAND 1")
  pipeline.call("COMMAND 2")
end
```

`redis-rb` offer some more convenience by having pipelined queries return a `Future` object as well.

```ruby
future_1, future_2
result_1, result_2 = client.pipelined do |pipeline|
  future_1 = pipeline.command1
  future_2 = pipeline.command2
end
future_1.value == result_1
future_2.value == result_2
[...]

PR #197, opened 2024-08-07
PoC: Pipelining support

casperisfine:
Ref: https://github.com/trilogy-libraries/trilogy/issues/189

This is the most basic API possible and uses a hack, but it demonstrate that it's possible to do pipelining with MySQL API.

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -937,6 +937,42 @@ static VALUE rb_trilogy_query(VALUE self, VALUE query)
     return execute_read_query_response(ctx);
 }
 
+static VALUE rb_trilogy_pipelined_query(VALUE self, VALUE queries)
+{
+    struct trilogy_ctx *ctx = get_open_ctx(self);
+
+    Check_Type(queries, T_ARRAY);
+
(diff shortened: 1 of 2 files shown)


contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -937,6 +937,42 @@ static VALUE rb_trilogy_query(VALUE self, VALUE query)
...
+        }
+
+        if (rc < 0) {
+            handle_trilogy_error(ctx, rc, "trilogy_query_send");
+        }
+    }
+
+    VALUE results = rb_ary_new2(count);
+    for (index = 0; index < count; index++) {
+        ctx->conn.packet_parser.sequence_number = 1; // HACK
```
casperisfine, line 968: I'm not sure if the sequence number can really be assumed to be always `1`. From what I could see it's the case, but we could also record the sequence number of each query.

composerinteralia: Won't be 1 if the query is long enough to span multiple MySQL packets.

casperisfine: Right, I think we need to store these values in an array somewhere to restore them.

PR #198, opened 2024-08-18
Add difference in execute method's response for trilogy and mysql2 to the documentation

khhizr007:
This pr has been created because when we switched to trilogy from mysql2 recently we faced an issue with the usage of ```ActiveRecord::Base.connection.execute```. We are using the response we get from this method call but the way we use this response breaks when we switch to trilogy from mysql2.

This is because of the difference in the response object we get when using **trilogy** and **mysql2**. This difference was not mentioned in the **mysql2 gem compatibility** section of the documentation. This commit adds these changes to the documentation to make it clear.

If there's any additional context that should be included here, I'd be happy to add it.

### DIFF
--- a/contrib/ruby/README.md
+++ b/contrib/ruby/README.md
@@ -86,3 +86,9 @@ differences:
   the transcoding step up to the caller.
 * There is no `as` query option. Calling `Trilogy::Result#each` will yield an array
   of row values. If you want a hash you should use `Trilogy::Result#each_hash`.
+* When using raw sql with the execute method of active_record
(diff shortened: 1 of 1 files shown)


composerinteralia:
Thank you for the contribution!

This section is for documenting notable differences between this gem and the mysql2 gem, not for documenting differences between the corresponding Active Record adapters. That documentation would belong somewhere in Rails, so I'm going to close this issue.

(FWIW I think execute with the Trilogy adapter is meant to return `Trilogy::Result`, not `Activerecord::Result`. It's possible that isn't quite right in Rails 7.2, but I think it works that way on Rails main. Although either way it's not expected to be 100% compatible with `Mysql2::Result`.)

Issue #192, opened 2024-07-05
Make C-extension Ractor-safe


mohamedhafez:
In planning ahead to the near future when TruffleRuby can run C-extensions marked with `rb_ext_ractor_safe(true)` in parallel, and for when Ractors are no longer just experimental, it would be great if the C-extension could be made Ractor-safe, or marked as such if it already is so!

mohamedhafez:
From what I hear, @nirvdrum has been running an app using this gem at Shopify on TruffleRuby without the C-extension lock, meaning its getting run concurrently in multiple threads, and hasn't had issues, so I'm _hoping_ there isn't much work to be done for this, and it just needs to be marked with `rb_ext_ractor_safe(true)`

jhawthorn:
I agree we should be able to set `rb_ext_ractor_safe(true)`. We do require that calls to a given Trilogy client are synchronized (done by the GVL) but I don't believe there is global state outside of them (so by Ractor rules, it should be safe).

Would you be interested in sending a PR?

mohamedhafez:
Sent in a quick pull request https://github.com/trilogy-libraries/trilogy/pull/200!

nirvdrum:
Sorry, I missed the ping on this. Yes, we've been running Trilogy on TruffleRuby for a project at Shopify and have done extensive load testing on it. TruffleRuby normally doesn't have a GVL, but imposes a lock for native extensions that don't declare they don't need one to avoid multi-threading issues there. It's possible to override that with the `--experimental-options --ruby.cexts-lock=false` options, which is how I ran it. We didn't see any issues running Trilogy that way over the course of hundreds of load tests.

This change will be really nice because TruffleRuby will know it doesn't need a lock when running Trilogy, which means that runtime option to override the default settings will not be needed.

PR #200, opened 2024-09-11
Mark as ractor-safe

mohamedhafez:
As per https://github.com/trilogy-libraries/trilogy/issues/192#issuecomment-2342174208, trilogy should already be ractor-safe, so marking it as such.


jhawthorn (approved):

jhawthorn:
Thank you!

PR #201, opened 2024-09-12
Remove mysql 5.7 from macos

jhawthorn:
Homebrew will no longer install mysql 5.7, so seems reasonable for us to drop it (at least for macos).


composerinteralia (approved):

composerinteralia:
👍🏻 I removed them from the required checks

PR #212, opened 2024-12-03
Speedup Trilogy#escape by 3 to 5x

byroot:
Copying characters one by one is very inneficient, especially since each time the buffer capacity has to be checked again and again.

Using `memcpy` for chunks of bytes that don't need any escaping, we can save a lot of time.

Before:

```
ruby 3.3.4 (2024-07-09 revision be1089c8ec) +YJIT [arm64-darwin23]
            2k-plain    23.802k i/100ms
       2k-few-escape    22.693k i/100ms
            2k-plain    239.746k (± 0.5%) i/s    (4.17 μs/i) -      1.214M in   5.063403s
       2k-few-escape    227.900k (± 0.6%) i/s    (4.39 μs/i) -      1.157M in   5.078454s
```

After:

```
            2k-plain    94.975k i/100ms
       2k-few-escape    61.998k i/100ms
            2k-plain      1.039M (± 1.0%) i/s -      5.224M in   5.025882s
       2k-few-escape    658.756k (± 1.3%) i/s -      3.348M in   5.082975s
```

Bench:

```ruby

require "trilogy"
require "benchmark/ips"
t = Trilogy.new(database: "test")

plain = "A" * 2000
few_escapes = (("a" * 19) + "\n") * 100
Benchmark.ips do |x|
  x.report("2k-plain") { t.escape(plain) }
  x.report("2k-few-escape") { t.escape(few_escapes) }
end
```

@jhawthorn @matthewd

### DIFF
--- a/inc/trilogy/buffer.h
+++ b/inc/trilogy/buffer.h
@@ -53,6 +53,19 @@ int trilogy_buffer_expand(trilogy_buffer_t *buffer, size_t needed);
  */
 int trilogy_buffer_putc(trilogy_buffer_t *buffer, uint8_t c);
 
+/* trilogy_buffer_write - Appends multiple bytes to the buffer, resizing the underlying
+ * allocation if necessary.
+ *
+ * buffer - A pointer to a pre-initialized trilogy_buffer_t.
+ * ptr    - The pointer to the byte array.
+ * len    - How many bytes to append.
+ *
+ * Return values:
+ *   TRILOGY_OK     - The character was appended to the buffer
+ *   TRILOGY_SYSERR - A system error occurred, check errno.
+ */
+int trilogy_buffer_write(trilogy_buffer_t *buffer, const uint8_t *ptr, size_t len);
+
 /* trilogy_buffer_free - Free an trilogy_buffer_t's underlying storage. The buffer
  * must be re-initialized with trilogy_buffer_init if it is to be reused. Any
  * operations performed on an unintialized or freed buffer are undefined.
--- a/src/client.c
+++ b/src/client.c
@@ -1,4 +1,5 @@
 #include <fcntl.h>
+#include <string.h>
 
 #include "trilogy/client.h"
 #include "trilogy/error.h"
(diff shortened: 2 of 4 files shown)


inc/trilogy/buffer.h
```diff
@@ -53,6 +53,19 @@ int trilogy_buffer_expand(trilogy_buffer_t *buffer, size_t needed);
...
+ *
+ * buffer - A pointer to a pre-initialized trilogy_buffer_t.
+ * ptr    - The pointer to the byte array.
+ * len    - How many bytes to append.
+ *
+ * Return values:
+ *   TRILOGY_OK     - The character was appended to the buffer
+ *   TRILOGY_SYSERR - A system error occurred, check errno.
+ */
+int trilogy_buffer_puts(trilogy_buffer_t *buffer, const uint8_t *ptr, size_t len);
```
jhawthorn, line 67: It's a nitpick, but `puts` to me implies a newline (like C's puts or Ruby's puts). Maybe we call this `trilogy_buffer_write`?

src/client.c
```diff
@@ -735,28 +736,40 @@ int trilogy_escape(trilogy_conn_t *conn, const char *str, size_t len, const char
...
     b->len = 0;
 
-    if (conn->server_status & TRILOGY_SERVER_STATUS_NO_BACKSLASH_ESCAPES) {
-        for (size_t i = 0; i < len; i++) {
-            const uint8_t c = (uint8_t)str[i];
+    // Escaped string will be at least as large as the source string,
+    // so might as well pre-expand the buffer.
+    CHECKED(trilogy_buffer_expand(b, len));
+
+    uint8_t *cursor = (uint8_t *)str;
```
jhawthorn, line 743: It's technically valid as-is, but we should be able to preserve str's `const`

```suggestion
    const uint8_t *cursor = (const uint8_t *)str;
```

src/client.c
```diff
@@ -735,28 +736,40 @@ int trilogy_escape(trilogy_conn_t *conn, const char *str, size_t len, const char
...
+    const uint8_t *end = cursor + len;
 
-            if (c == '\'') {
-                CHECKED(trilogy_buffer_putc(b, '\''));
-                CHECKED(trilogy_buffer_putc(b, '\''));
+    if (conn->server_status & TRILOGY_SERVER_STATUS_NO_BACKSLASH_ESCAPES) {
+        while (cursor < end) {
+            uint8_t *next_backslash = memchr(cursor, '\'', (size_t)(end - cursor));
```
jhawthorn, line 748: This isn't actually the next backslash, it's the next quote. Maybe we call this `next_escape`?

jhawthorn:
Looks great to me other than minor naming issues!

jhawthorn (approved):

jhawthorn:
Thank you!

PR #214, opened 2024-12-14
ruby: Implement a buffer pool

byroot:
The trilogy client eagerly allocate a 32kiB buffer, and grows it as needed. It's never freed not shrunk until the connection is closed. Since by default the MySQL `max_allowed_packet` is 16MiB, long living connections will progressively grow to that size.

For basic usage it's not a big deal, but some applications may have dozens if not hundreds of connections that are mostly idle.

A common case being multi-tenant applications with horizontal sharding. In such cases you only ever query one database but have open connections to many databases.

This situation might lead to a lot of memory retained by trilogy connections and never really released, looking very much like a memory leak.

This can be reproduced with a simple script:

```ruby
require 'trilogy'
connection_pool = []

50.times do
  t = Trilogy.new(database: "test")
  t.query("select '#{"a" * 16_000_000}' as a")
  connection_pool << t
end

puts "#{`ps -o rss= -p #{$$}`} kiB"
```

```
$ ruby /tmp/trilogy-leak.rb
927120 kiB
```

If we instead take over the buffer lifetime management, we can implement some pooling for the buffers, we can limit the total number of buffer to as many connections are actually in use concurrently.

The same reproduction script with the current branch:

```
$ ruby -Ilib:ext /tmp/trilogy-leak.rb
108144 kiB
```

NB: The implementation isn't quite ideal, as the buffers are allocated by the Ruby extension with `ruby_xmalloc`, but grown by trilogy itself when needed with raw `realloc`. It works but ideally there should be some hook in the `trilogy` C library so we can use the ruby allocator all the way. I can try to work on that if it's really a blocker

Also cc @shanempope this is very likely to explain the memory growth observed last month after reforking stops, because:
  - Active Record closes all DB clients on fork, so as long as the app reforks, this "leak" is kept in check.
  - When we hit OOM, we do a heap dump from a forked child, so `Trilogy` here again, Active Record close all clients and free the buffers. Making the source of the growth disappear.

All this to say you probably want to experiment with this branch or something like it.

@adrianna-chang-shopify @jhawthorn @matthewd


byroot:
The macOS build seem to fail consistently, but It doesn't contain any usable info and I can't reproduce it locally :/

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -15,6 +15,161 @@
...
+
+    if (pool->len >= buffer_pool_max_size) {
+        xfree(buffer->buff);
+        buffer->buff = NULL;
+        buffer->cap = 0;
+        return false;
+    }
+
+    if (!pool->capa) {
+        pool->entries = RB_ALLOC_N(buffer_pool_entry, 16);
```
adrianna-chang-shopify, line 143: If I'm understanding correctly, we set the default `buffer_pool_max_size` to be 8, but then allocate 16 entries here when initializing a new pool. Wouldn't it make more sense to allocate 8 entries to start with (or to allocate the configured max size?) Why 16 specifically?

byroot:
> Is this still the case? I see us using `RB_ALLOC_N` when initializing the pool, and then `RB_REALLOC_N` when resizing.

That's for the pool itself. The buffers inside the pool are reallocated from `buffer.c` in the C library: https://github.com/trilogy-libraries/trilogy/blob/c080567a95f7c412b9bffffbabf59a549fa87e5e/src/buffer.c#L39

So there is a mix of regular `malloc/free/realloc` and the ruby versions of those. Which works most of the time but is kinda incorrect and would crash if Ruby is built with some specific debug flags on.

contrib/ruby/ext/trilogy-ruby/cext.c
byroot: Yes, it's a bit of a difference. Those should probably match. 

Initially I had some code that would dynamically size the pool based on the number of live `Trilogy` objects, but that caused issues with the GC (since the pool is retained by an object, when both the pool and the client are GCed, the pool may be freed before the clients). Hence why I fell back to just a global setting like this.

I started with 16, but realized that `16 * 16MiB` is `256 MiB`, so figured 8 was already plenty.

jhawthorn:
Mixing the xmalloc and malloc has actually caused us significant issues in terms of causing extra GC due to the accounting "leak". We do test in our CI (occasionally) to ensure that they aren't mismatched.

byroot:
> Mixing the xmalloc and malloc has actually caused us significant issues in terms of causing extra GC

Makes sense.

I suppose there's two way to fix that:
  - Refactor even more heavily
  - Just alloc and free those buffer with regular malloc from `contrib/ruby`

Do you have a preference?

byroot:
Alright, I did the later and that fixed the macOS build, so yeah, that indeed wasn't a good idea.

adrianna-chang-shopify (approved):

igrigorik:
FYI, a _much_ simpler strategy that we tried and proved to perform well:
https://github.com/igrigorik/trilogy/commit/9120d5d6866bee4a489b27bec491d357965e54aa

byroot:
It may be much simpler but is also much less effective.

igrigorik:
@byroot can you elaborate on the "much less effective"? At least for our use case, we observe that most queries rarely hit MB+, so downsizing after occasional "fat row" is not a common occurrence and overall memory footprint is contained even with the simple distinct buffer approach.

byroot:
It's much less effective because, you have to wait for a small query to be executed on the connection before you shrink the buffer, which in sharding scenario could be a long time. So first it's not making the memory immediately available, but eventually later.

It's much less efficient, because with your patch, the buffers can constantly "yo-yo" between multiple sizes, requiring lots of `realloc`, meanings lots of byte copying. Also this frequent `realloc` could potentially cause fragmentation depending on the allocator (probably fine for `jemalloc`). 

I absolutely don't doubt it's enough to keep memory usage in check for your use case, but that doesn't make it a good general solution, it's just a quick targeted hack. You can run the repro posted above with both branch and see the difference.

byroot:
Thanks. I think that's probably caused on exit by the lazily allocated per-ractor pool. I think it can be fixed though, I'll have a look tomorrow.

byroot:
I need to run, but I found a good repro:

```ruby
require "trilogy"

t = Thread.new do
  t = Trilogy.new
  p t.query("select sleep(10)")
end
t.join(0.1)
```

rafaelfranca:
Thanks Jean!

byroot:
@rafaelfranca 3a798420b7d7685a96766e7b589e646a11321383 fixes the bug and adds a test case to properly cover it.

paracycle:
This PR has been running on Shopify Storefront Renderer in production for about a month now with no problems.

I think we can merge and make a release with this. /cc @jhawthorn and @eileencodes

jhawthorn:
Thank you!

I removed the configurable buffer pool size in an additional commit

PR #215, opened 2024-12-14
Fix Ractor support

byroot:
Ref: https://github.com/trilogy-libraries/trilogy/issues/192

The intent was more to allow TruffleRuby not to use a lock, but still, I find it a bit weird that the C extension declare ractor safety and yet isn't usable from a Ractor.


jhawthorn (approved):

jhawthorn:
Thanks!

PR #219, opened 2025-02-05
feat: use precompiled so/a library to build with it a ruby source extension

malbajun:
- [x] * use precompiled so/a library to build with it a ruby source extension


jhawthorn:
Sorry, I don't see the motivation here. By design the library code is plenty small enough to just be compiled with the gem.

PR #220, opened 2025-02-05
feat: build with cmake

malbajun:
- [x] + build with cmake


jhawthorn:
Thanks, but we have no interest in changing the build system

PR #223, opened 2025-02-18
Add treatment for proxied ssl connection error

pcasaretto:
This error popped up on our exception tracker:

```
Trilogy::ProtocolError: 2026: TLS/SSL error: Connection timed out (110) (trilogy_query_recv)
```

Rails wrapped it as a `ActiveRecord::StatementInvalid`.

Although [2026 is client error](https://dev.mysql.com/doc/mysql-errors/8.0/en/client-error-reference.html#error_cr_ssl_connection_error) and trilogy itself has no reference to it, when it is used in tandem with a proxy like proxysql that is both a client and a server the error can surface.

I'm proposing wrapping this error with a BaseConnectionError.

This seems to be inline with what mysql2 does https://github.com/brianmario/mysql2/pull/911/files#diff-2c27fbaef84b063b17b3eb069adf94acd254bbbf7e197073e0d2533d8d0441f1R39

### DIFF
--- a/contrib/ruby/Gemfile.lock
+++ b/contrib/ruby/Gemfile.lock
@@ -11,6 +11,7 @@ GEM
     bigdecimal (3.1.8)
     minitest (5.25.4)
     mysql2 (0.5.6)
+    openssl (3.3.0)
     rake (13.0.1)
     rake-compiler (1.0.7)
       rake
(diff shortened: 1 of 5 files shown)


pcasaretto:
After discussing it with a colleague, we've agreed that it wouldn't make sense for Trilogy to have proxysql specific error treatment.

PR #226, opened 2025-04-01
ruby: Prevent concurrent usage of Trilogy connections

casperisfine:
Using a trology connection in a concurrent way will likely lead to various protocol errors, but can also lead to various crashes.

As such it is preferable if Trilogy raises a clear to understand and deterministic error.

I chose to implement this on the Ruby side, because it's simpler and more straightforward.

I initially tried to do it in C, but the `_cb_wait` callback may raise or throw, so we'd need to use `rb_protect`, and release the lock there. The problem is the callback doesn't have access to the `trilogy_context` struct, so can't release the lock.

If we're adament this should be done in C, we'll need a much larger refactor.

cc @tenderlove @jhawthorn

### DIFF
--- a/contrib/ruby/lib/trilogy.rb
+++ b/contrib/ruby/lib/trilogy.rb
@@ -7,7 +7,27 @@ require "trilogy/cext"
 require "trilogy/encoding"
 
 class Trilogy
+  Synchronization = Module.new
+
+  source = public_instance_methods(false).map do |method|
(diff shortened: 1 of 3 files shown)


samuel-williams-shopify:
This makes sense to me! Nice work. I don't have a strong opinion about the implementation, but this is better than crashing for sure.

Issue #228, opened 2025-05-06
Cannot compile native extension on Fedora 42 w/ GCC 15


rightpad:
Hello,

I'm using Fedora Linux w/ rbenv to manage rubies; everything has been great until I recently upgraded to Fedora 42 with GCC 15. Due to incompatibilities with other gems (https://github.com/RubyCrypto/ed25519/pull/45) and a related issue on the ruby issue tracker (https://bugs.ruby-lang.org/issues/21290), I suspect the compiler upgrade is responsible for the errors I'm seeing when attempting to compile trilogy during a routine gem install (I tried both rubies 3.4.3 & 3.4.2). I've attached the log files here. Unfortunately I'm not well versed in the C side of Ruby so I can't help with a pull request, but I'm ready to help test any solutions if needed.

[gem_make.out.log](https://github.com/user-attachments/files/20071059/gem_make.out.log)

[mkmf.log](https://github.com/user-attachments/files/20071061/mkmf.log)

composerinteralia:
Looks like a fix was provided in Ruby and should be available in the next patch releases: https://github.com/trilogy-libraries/trilogy/pull/229#issuecomment-2899295546

PR #229, opened 2025-05-15
Fix building gem with gcc 15.1.1

skipkayhil:
Fixes https://github.com/trilogy-libraries/trilogy/issues/228

```
Installing trilogy 2.9.0 with native extensions
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    current directory: /home/hartley/.local/share/mise/installs/ruby/3.4.3/lib/ruby/gems/3.4.0/gems/trilogy-2.9.0/ext/trilogy-ruby
/home/hartley/.local/share/mise/installs/ruby/3.4.3/bin/ruby extconf.rb
checking for CRYPTO_malloc() in -lcrypto... *** extconf.rb failed ***
Could not create Makefile due to some reason, probably lack of necessary
libraries and/or headers.  Check the mkmf.log file for more details.  You may
need configuration options.
```

Previously, trying to build Trilogy with gcc 15.1.1 would result in an error like above. Other gems with c extensions have encountered the same error with gcc 15.1.1, and their fix was to use `append_cflags` instead of modifying `$CFLAGS`. (https://github.com/RubyCrypto/ed25519, https://github.com/tonytonyjan/jaro_winkler/pull/62)

Testing locally, `bundle exec rake build && gem install pkg/*.gem` fails with the above error before this patch but is able to install the gem after.

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/extconf.rb
+++ b/contrib/ruby/ext/trilogy-ruby/extconf.rb
@@ -10,7 +10,8 @@ File.binwrite("trilogy.c",
   }.join)
 
 $objs = %w[trilogy.o cast.o cext.o]
-$CFLAGS << " -I #{__dir__}/inc -std=gnu99 -fvisibility=hidden"
+
+append_cflags(["-I #{__dir__}/inc", "-std=gnu99", "-fvisibility=hidden"])
 
 dir_config("openssl")
 


skipkayhil:
Fixed by rebuilding my Ruby with https://github.com/ruby/ruby/commit/2297afda7ff3926c51fea700dfbf0f0eb4fea1e5

(I used `MISE_RUBY_APPLY_PATCHES=https://github.com/ruby/ruby/commit/2297afda7ff3926c51fea700dfbf0f0eb4fea1e5.patch mise install ruby 3.4.4`)

PR #231, opened 2025-07-03
Fix native extension build with GCC 15

davue:
**OS:** Fedora 42
**Ruby:** 3.2.1
**GCC:** 15.1.1

The native extension build is broken with GCC 15 resulting in the same errors as described here: https://bugs.ruby-lang.org/issues/21290

CFLAGS should be appended with `append_cflags` instead of directly appending them which fixes the problem.

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/extconf.rb
+++ b/contrib/ruby/ext/trilogy-ruby/extconf.rb
@@ -10,7 +10,7 @@ File.binwrite("trilogy.c",
   }.join)
 
 $objs = %w[trilogy.o cast.o cext.o]
-append_cflags(["-I #{__dir__}/inc","-std=gnu99","-fvisibility=hidden"])
+append_cflags(["-I #{__dir__}/inc", "-std=gnu99", "-fvisibility=hidden"])
 
 dir_config("openssl")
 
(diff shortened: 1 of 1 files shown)


davue:
Duplicate of #229

davue:
I cannot reopen the older PR because it is not mine so I'm tracking this here.

I still think we should merge this for two reasons:
- It is generally recommended to use `append_cflags` instead of manually appending it (or is there a reason it was done like this here?)
- To support Ruby versions < 3.4.5

contrib/ruby/ext/trilogy-ruby/extconf.rb
```diff
@@ -10,7 +10,7 @@
   }.join)
 
 $objs = %w[trilogy.o cast.o cext.o]
-$CFLAGS << " -I #{__dir__}/inc -std=gnu99 -fvisibility=hidden"
+append_cflags(["-I #{__dir__}/inc","-std=gnu99","-fvisibility=hidden"])
```
composerinteralia, line 13: spacing nitpick:

```suggestion
append_cflags(["-I #{__dir__}/inc", "-std=gnu99", "-fvisibility=hidden"])
```

composerinteralia (approved):
Using `append_cflags` makes sense to me and matches what the docs recommend: https://docs.ruby-lang.org/en/master/extension_rdoc.html#label-Prepare+extconf.rb

PR #238, opened 2025-08-04
Test mysql 8.4

composerinteralia:


issyl0:
Looks like `8.4` and `8.4-oracle` are the only published ones. There's no `-debian` tags since `8.0`.

composerinteralia:
These images use `oracle-linux:9-slim` https://github.com/docker-library/mysql/blob/4ea634473f3ca9b1dcd476236e6d617258e2d63b/8.0/Dockerfile.oracle#L7. I'm having trouble figuring out where to get community-mysql-test from now.

test/mysql/8.4-conf.d/build.cnf
```diff
@@ -0,0 +1,21 @@
+# This MySQL configuration file is mounted into the Database container (/etc/mysql/conf.d)
```
ngan, line 1: Probably just nit, but I would structure this as `test/mysql/conf.d/{MYSQL_VERSION}/build.cnf` then mount `test/mysql/conf.d/{MYSQL_VERSION}`

PR #251, opened 2025-11-24
Fix use-after-close race causing SEGV

jeremy:
Fixes a race condition that causes SIGSEGV when a connection is closed while another thread is mid-query.

The library isn't meant to be thread-safe, but in case of mistaken concurrent use let's raise exceptions rather than crashing. Prefer both belt and suspenders when they're cheap and available.

Adds NULL deref guards to protect against connection close while awaiting IO (rb_wait_for_single_fd without GVL). Now raises Trilogy::ConnectionClosed instead of segfaulting.

Guards:
* read_packet: check socket and buffer before read
* trilogy_flush_writes: check socket before write
* current_packet_type: check buffer before access
* Ruby wrapper flush_writes: early exit on closed socket
* Ruby read_query_response: check socket after wait returns
* Ruby buffer_checkout/checkin: NULL pool fallback

/cc @jhawthorn

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -119,7 +119,8 @@ static inline buffer_pool *get_buffer_pool(void)
 static void buffer_checkout(trilogy_buffer_t *buffer, size_t initial_capacity)
 {
     buffer_pool * pool = get_buffer_pool();
-    if (pool->len) {
+
+    if (pool && pool->len) {
         pool->len--;
         buffer->buff = pool->entries[pool->len].buff;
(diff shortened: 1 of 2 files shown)


jhawthorn:
I don't think I agree with this approach. Using a connection in parallel is so dangerous that a SEGV is preferable to a silent failure (or a failure that may be retried at another level of the stack, like `TRILOGY_CLOSED_CONNECTION` is likely to be)

jeremy:
Agreed the preconditions and checks are inherently racy and are incomplete without synchronization. I'll withdraw until that's addressed and these can become invariant checks (indicating improper synchronization impl) rather than concurrency bomb sniffers.

PR #264, opened 2026-01-29
Implement Trilogy#abandon_results!

byroot:
API is similar to MySQL2, it is functionally equivalent to `client.next_result while client.more_results_exist?` but saves having to parse and allocate the responses.


byroot:
> FWIW #40

Interesting.

I honestly doubt:

> "cheaper" to just throw away (close) the connection and open a new one

Do hold true, especially when SSL is involved, and it's also costly for the server, and you then need several roundtrips to configure the connection.

byroot:
Either way, the reason I implemented this is that Rails trilogy adapter does the looping thing: https://github.com/rails/rails/blob/90ccb4501a2eecb85ee8978994034941d94ebf31/activerecord/lib/active_record/connection_adapters/trilogy/database_statements.rb#L52-L54

It probably isn't a big deal as I assume most queries don't have extra results, but seems cleaner.

PR #271, opened 2026-03-18
Replace sscanf datetime parsing with Go-inspired byte-arithmetic

schlubbi:
## Summary

Replace all three `sscanf()`-based datetime parsing blocks (DATETIME/TIMESTAMP, DATE, TIME) in `cast.c` with inline byte-arithmetic parsers inspired by Go's [`go-sql-driver/mysql` `parseDateTime()`](https://github.com/go-sql-driver/mysql/blob/master/utils.go#L108-L228) implementation, **and** replace `rb_funcall(rb_cTime, :utc, 7, ...)` Time object construction with Ruby's C-level `rb_time_timespec_new()` API.

### sscanf → byte-arithmetic parsing

`sscanf()` is a general-purpose format parser with overhead from format string interpretation, locale handling, and variadic argument processing. Go's MySQL driver avoids this entirely by operating directly on the byte buffer with simple arithmetic (`byte - '0'`), using length-based dispatch to determine the datetime format.

### rb_funcall → rb_time_timespec_new (per [byroot's suggestion](https://github.com/trilogy-libraries/trilogy/pull/271#issuecomment-4081623944))

`rb_funcall(rb_cTime, :utc, 7, ...)` goes through Ruby method dispatch, creates 7 `INT2NUM` VALUE arguments, unpacks them back to C integers in `time_arg()`, and converts to epoch via `timegmw()`. [`rb_time_timespec_new()`](https://github.com/ruby/ruby/blob/45dbc5a4a24cac30771b8c8353abbbfd35fa86b8/include/ruby/internal/intern/time.h) skips all of that — we compute the epoch in C and hand it directly to the allocator.

For UTC epoch computation, uses [Howard Hinnant's `civil_to_days` algorithm](http://howardhinnant.github.io/date_algorithms.html) (the foundation of C++20 `std::chrono::sys_days`) which handles the full MySQL 1000–9999 year range. POSIX `timegm()` was considered but fails for `tm_year < 0` on macOS; Ruby's own `timegm_noleapsecond()` handles this but is `static` internal.

### Byte-arithmetic parsing

New helpers in `cast.c` inspired by Go's [`bToi`](https://github.com/go-sql-driver/mysql/blob/master/utils.go#L225-L228), [`parseByte2Digits`](https://github.com/go-sql-driver/mysql/blob/master/utils.go#L212-L223), [`parseByteYear`](https://github.com/go-sql-driver/mysql/blob/master/utils.go#L199-L210), [`parseByteNanoSec`](https://github.com/go-sql-driver/mysql/blob/master/utils.go#L182-L197):

| Function | Purpose | Go equivalent |
|----------|---------|---------------|
| `byte_to_digit(b)` | Single ASCII char → int | `bToi()` |
| `parse_2digits(p)` | 2-byte parse (month, day, hour, min, sec) | `parseByte2Digits()` |
| `parse_4digits(p)` | 4-byte year parse | `parseByteYear()` |
| `parse_microseconds(p, len)` | 1-6 fractional digits → microseconds | `parseByteNanoSec()` |

Replaced parsing blocks:
- **DATETIME/TIMESTAMP**: Length-based dispatch (10/19/21-26 bytes) with direct byte access at known offsets
- **DATE**: Exact 10-byte parse
- **TIME**: 8-15 byte parse with fractional seconds

### C-level Time construction

| Before | After |
|--------|-------|
| `rb_funcall(rb_cTime, id_utc, 7, INT2NUM(y), ...)` | `rb_time_timespec_new(&ts, INT_MAX-1)` |
| Method dispatch + 7 VALUE args + time_arg() + timegmw() | Direct epoch → Time allocation |

- `civil_to_epoch_utc()` — Hinnant algorithm for UTC (portable, no timegm)
- `mktime()` — standard C for local time with DST handling
- `trilogy_make_time()` — unified helper for both DATETIME and TIME paths
- Removed unused `id_local`, `id_localtime`, `id_utc`

### Tests

20 new test cases covering:
- Fractional seconds with 1-6 digits of precision (Go-ported)
- Zero date/datetime handling
- Specific date and datetime parsing validation
- TIME column with microsecond precision
- **Epoch edge cases**: Unix epoch zero, pre-1970 dates, leap year Feb 29, non-leap century (1900), far future (9999), MySQL minimum year (1000), leap year with local timezone

## Profile Results

CPU-time isolated benchmark (per-row incremental casting cost, GC disabled):

```
Original (sscanf + rb_funcall):    ~208 ns/row
After byte-arithmetic parsing:      ~75 ns/row  (2.7× faster)
After rb_time_timespec_new:          ~3 ns/row  (72× faster)
```

The casting overhead is now indistinguishable from no-casting in `benchmark-ips` measurements — all datetime benchmarks report "same-ish: difference falls within error."

## Files

| File | Change |
|------|--------|
| `contrib/ruby/ext/trilogy-ruby/cast.c` | Replace sscanf + rb_funcall with byte-arithmetic + C-level Time API |
| `contrib/ruby/test/cast_test.rb` | Add 20 test cases (Go-ported + epoch edge cases) |

## Test Results

```
45 runs, 164 assertions, 0 failures, 0 errors, 0 skips
```

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cast.c
+++ b/contrib/ruby/ext/trilogy-ruby/cast.c
@@ -80,10 +80,6 @@ static void cstr_from_value(char *buf, const trilogy_value_t *value, const char 
     buf[value->data_len] = 0;
 }
 
-// Build a Ruby Time object via rb_time_timespec_new (C API) instead of
-// rb_funcall(rb_cTime, :utc/:local, 7, ...) to skip method dispatch,
-// 7x INT2NUM, and Ruby's internal time_arg() + timegmw() conversion.
-//
 // For UTC: uses the Hinnant civil_to_days algorithm (C++20 std::chrono
 // foundation, handles the full MySQL 1000-9999 year range without timegm).
 // For local: uses mktime (standard C) which consults the system timezone.


byroot (commented):
I like that change a lot.

I think we should just remove a bit of boiler plate from the tests, and I'm not convinced it's really worth committing the benchmark. But other than that 👍 

@composerinteralia any opinions?

schlubbi:
> I like that change a lot.
> 
> I think we should just remove a bit of boiler plate from the tests, and I'm not convinced it's really worth committing the benchmark. But other than that 👍
> 

🎈 
Bumped the precision of the existing datetime/time columns and am reusing them in the tests now. Benchmark script is gone from the PR too.

PR #293, opened 2026-05-11
cext.c: Copy FD flags after duping

### LINKED ISSUE #292: Trilogy leak file descriptors on exec
Reported to me by @dannyfallon.

I haven't yet had time to dig into the code to find the root cause, but I had time to write a reproducer:

```ruby
gem "bigdecimal"
require "trilogy"

count = Integer(ENV["COUNT"] || 100)

conn = Trilogy.new
conn.ping
pipe = IO.pipe.last
p pipe.fileno

if count > 0
  ENV["COUNT"] = (count - 1).to_s
  Process.exec("ruby", $0, close_others: true)
end
```

Expected:

It should print a consitent low number 100 times.

Actual:

The pipe file descriptor keep incrementing, showing that after each `exec` we have one more leaked file descriptor.

### Hunch

At first I wouldn't believe it, because `Trilogy` create the TCP socket using Ruby's `TCPSocket.new`, which does set `CLOEXEC` so these FD should be closed.

But I suspect we're duping the FD in C, or something like that, and are probably losing the flag there.

byroot:
Fix: https://github.com/trilogy-libraries/trilogy/issues/292

`dup(2)` doesn't carry any flags over, meaning we lose important flags like `FD_CLOEXEC`.

NB: still need to add a test

cc @dannyfallon.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+### Fixed
+
+- Properly set `FD_CLOEXEC` on trilogy sockets created by Ruby to avoid leaking file descriptor on `exec`.
+
 ## 2.12.4
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


byroot:
I think the ideal fix would be not to dup at all, and instead set `.autoclose=false` on the Ruby IO object, but it need to be done carefully to ensure we'd never raise without explictly closing the fd.

PR #307, opened 2026-07-30
Default `ssl_mode` to `SSL_PREFERRED_NOVERIFY` for TCP connections

### LINKED ISSUE #306: Prefer TLS by default when `ssl_mode` is omitted
## Summary

I'd like to propose changing Trilogy's default `ssl_mode` to `PREFERRED_NOVERIFY` when `ssl_mode` is not explicitly configured.

This would align Trilogy's default behavior with MySQL client programs and reduce surprises when migrating from `mysql2`.

## Motivation

Currently, Trilogy establishes an unencrypted connection by default, even when the server supports TLS.

In contrast, MySQL client programs (`libmysqlclient`) default to `SSL_MODE_PREFERRED`, attempting a TLS connection first and falling back to an unencrypted connection only if TLS is unavailable.

As a result, applications migrating from `mysql2` to Trilogy may unexpectedly switch from encrypted connections to unencrypted connections without changing their configuration.

This difference is becoming more noticeable as managed MySQL services (for example, Aurora MySQL 8.4 by default) increasingly expect clients to use TLS.

## Proposed behavior

When `ssl_mode` is omitted:

- Attempt a TLS connection.
- Fall back to an unencrypted connection if TLS is unavailable.
- Continue respecting any explicitly configured `ssl_mode`.

I'd be happy to prepare a PR with tests and documentation if this proposal sounds reasonable.

## References

- MySQL 8.4 Reference Manual – Using Encrypted Connections  
  https://dev.mysql.com/doc/refman/8.4/en/using-encrypted-connections.html

- MySQL 8.4 Reference Manual – `--ssl-mode` (`SSL_MODE_PREFERRED` is the default)  
  https://dev.mysql.com/doc/refman/8.4/en/connection-options.html#option_general_ssl-mode

- Amazon Aurora MySQL – Security (`require_secure_transport=ON` by default in Aurora MySQL 8.4)  
  https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/AuroraMySQL.Security.html

Yuhi-Sato:
Fixes #306

## Summary

When the `ssl_mode` option is omitted, the Ruby binding now defaults TCP connections to `Trilogy::SSL_PREFERRED_NOVERIFY`: TLS is attempted, and the connection falls back to plaintext when the server doesn't support it. This matches the default behavior of MySQL client programs / `libmysqlclient` (`SSL_MODE_PREFERRED`) and avoids silently downgrading to plaintext when migrating from `mysql2`.

- The server certificate is **not** verified in this mode; pass `SSL_VERIFY_CA` / `SSL_VERIFY_IDENTITY` explicitly when verification is needed, or `SSL_REQUIRED_NOVERIFY` to make TLS mandatory.
- An explicit `ssl_mode` is honored as before, including `SSL_DISABLED`.
- Unix socket connections still default to plaintext, matching `libmysqlclient`'s PREFERRED behavior of not encrypting them.
- Ruby binding only — no changes to the C API defaults, the `trilogy_ssl_mode_t` enum, or `trilogy_sockopt_t` zero-value semantics.

## Tests

Adds `DefaultSslModeTest`, separate from `SslTest` because `SslTest#setup` skips everything on non-TLS servers while the fallback test here must run exactly then:

- omitted `ssl_mode` connects with TLS on a TLS-capable server
- explicit `SSL_DISABLED` stays plaintext (guards the zero value against being clobbered by the default, since `SSL_DISABLED == 0`)
- omitted `ssl_mode` falls back to plaintext on a server without TLS — this runs on the MariaDB 10.6/10.11 CI jobs, which don't enable TLS
- Unix socket connections stay plaintext by default

The tests call `Trilogy.new` directly because the `new_tcp_client` helper passes `ssl_mode` explicitly, and check `SHOW SESSION STATUS LIKE 'Ssl_cipher'` rather than `performance_schema.session_status` because the fallback test runs on servers where `performance_schema` is OFF (the MariaDB < 11.4 default) and that table is empty.

## Docs

- README: new SSL/TLS section describing the default, the fallback, the lack of certificate verification, and how to opt into stricter modes.
- CHANGELOG entry under Unreleased.


composerinteralia (commented):
Changing the default seems OK to me, but might need to wait for a major release since it's a breaking change.

Yuhi-Sato:
@composerinteralia 
I understand, thanks!

byroot:
> since it's a breaking change.

Is it though? I may be missing something, but since it automatically fall backs to unencrypted (the current default), I wouldn't expect breakages.

Unless somehow using unverified SSL instead of plain text cause some ripple effect.

composerinteralia:
Ah yeah, good point. I guess it's not breaking then 👍🏻

Yuhi-Sato:
@byroot 
I agree it probably won't cause breakages because of the fallback. 

My only concern is that it changes the default behavior, and while the SSL overhead is probably negligible, it's not entirely free.

byroot:
We can bump the major, but the annoying thing is that active record won't boot https://github.com/rails/rails/blob/0f36bbf72cc8b814bf1ad05c896c9c427b18217f/activerecord/lib/active_record/connection_adapters/trilogy_adapter.rb#L5

Yuhi-Sato:
@byroot 
Ah, I see. Since I also opened #309, I think a major version bump makes sense. I'd be happy to follow up with a Rails PR to bump the supported Trilogy version once Trilogy 3 is released.

PR #308, opened 2026-07-30
Use SHOW SESSION STATUS for SSL status queries in tests

Yuhi-Sato:
## Summary

`SslTest` checks TLS state via `performance_schema.session_status`, which is empty when the server runs with `performance_schema = OFF` — MariaDB's default in all versions. Since MariaDB 11.4 enables TLS out of the box (`have_ssl = YES`), `SslTest` runs but errors against any unconfigured MariaDB 11.4+ server, such as a stock brew or Docker install:

```
Expected: [["Ssl_version", "TLSv1.2"]]
  Actual: []
```

CI only passes because its MariaDB 11.4/11.8 configs switch `performance_schema` on.

`SHOW SESSION STATUS LIKE '...'` returns the same rows regardless of server tuning, so this PR uses it instead and drops the now-unneeded `performance_schema = ON` configs:

- rewrite the 9 `performance_schema.session_status` queries in `ssl_test.rb` (assertions unchanged — the result shape is identical)
- remove `performance_schema = ON` from `test/mariadb/conf.d/11.4` and `11.8` `build.cnf`
- delete `test/mariadb/conf.d/11.4/macos.cnf` and `11.8/macos.cnf`, which contained only that setting

## Verification

- stock `mariadb:11.4` (`have_ssl = YES`, `performance_schema = OFF`): 6 failures before, 0 after
- stock `mysql:8.0` (matches CI, `performance_schema` ON by default): 20 runs, 0 failures — no behavior change
- full `bundle exec rake test` run: 0 failures


composerinteralia (approved):
Less configuration seems good to me 👍🏻. Thank you

PR #309, opened 2026-07-30
Default `connect_timeout` to 5 seconds

Yuhi-Sato:
## Summary

When the `connect_timeout` option is omitted, connecting can currently block indefinitely: `TCPSocket.new` is called with `connect_timeout: nil`, and the C-side handshake wait falls back to `write_timeout`, which is also unlimited by default. The Ruby binding now defaults `connect_timeout` to 5 seconds (originally proposed as 120 seconds to match `mysql2`, lowered to 5 per review feedback).

- The default is applied inline when creating the TCP socket in `Trilogy#initialize`; the passed options hash and `connection_options` are not modified.
- It is only applied when the option is absent: passing `connect_timeout: nil` explicitly disables the timeout and restores the previous behavior, same semantics as `mysql2`.
- An explicit `connect_timeout` value is honored as before.
- Ruby binding only — no changes to the C API defaults or `trilogy_sockopt_t` zero-value semantics.

## Tests

- an omitted `connect_timeout` times out after the 5s default when connecting to a blackholed address (with an upper bound to distinguish it from the OS-level timeout)
- `test_connect_timeout_with_only_write_timeout` now passes `connect_timeout: nil` explicitly — otherwise the default would mask the `write_timeout` fallback path it exercises

## Docs

- README: note the 5s default and the `connect_timeout: nil` opt-out in the "mysql2 gem compatibility" section, since it differs from mysql2's 120s.
- CHANGELOG entry under Unreleased.

### DIFF
--- a/contrib/ruby/lib/trilogy.rb
+++ b/contrib/ruby/lib/trilogy.rb
@@ -42,8 +42,8 @@ class Trilogy
 
   def initialize(options = {})
     options[:port] = options[:port].to_i if options[:port]
-    # Set default connect_timeout, matching mysql2, to avoid hanging indefinitely on connect
-    options[:connect_timeout] = 120 unless options.key?(:connect_timeout)
+    # Set default connect_timeout to avoid hanging indefinitely on connect
+    options[:connect_timeout] = 5 unless options.key?(:connect_timeout)
     mysql_encoding = options[:encoding] || "utf8mb4"
     encoding = Trilogy::Encoding.find(mysql_encoding)
     charset = Trilogy::Encoding.charset(mysql_encoding)
--- a/contrib/ruby/lib/trilogy.rb
+++ b/contrib/ruby/lib/trilogy.rb
@@ -42,8 +42,6 @@ class Trilogy
 
   def initialize(options = {})
     options[:port] = options[:port].to_i if options[:port]
-    # Set default connect_timeout to avoid hanging indefinitely on connect
-    options[:connect_timeout] = 5 unless options.key?(:connect_timeout)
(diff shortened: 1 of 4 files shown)


byroot:
This still seems insanely high to me. I understand the logic of re-using `mysql2` default, but IMO this is a bad one.

If you need 2 minutes to connect you are having a huge problem. Something like 5 seconds would make sense. @composerinteralia thoughts?

composerinteralia:
I agree that 120 is overly generous and 5 or so seems reasonable. And even then most folks should probably lower it.

We could also consider a low default for the `write_timeout` (but not the `read_timeout`)

Yuhi-Sato:
That makes sense. I’ll update the default connect_timeout to 5 seconds.

A low default for write_timeout also seems worth considering, although I think it may be better handled separately since it could affect large writes and other legitimate workloads differently from connection establishment.

contrib/ruby/lib/trilogy.rb
byroot:
```suggestion
        connect_timeout = options.fetch(:connect_timeout, 5) || options[:write_timeout]
```

contrib/ruby/lib/trilogy.rb
```diff
@@ -42,6 +42,8 @@ def initialize(...)
 
   def initialize(options = {})
     options[:port] = options[:port].to_i if options[:port]
+    # Set default connect_timeout to avoid hanging indefinitely on connect
+    options[:connect_timeout] = 5 unless options.key?(:connect_timeout)
```
byroot:
```suggestion
```

byroot (approved):

PR #313, opened 2026-08-05
rb_trilogy_connect: add GC guard for all string configs

### LINKED ISSUE #312: Connection error message formatted from a dangling char* after try_connect (GC compaction)
## Summary

`Trilogy#initialize` takes a `char *` into the `host`/`socket` Ruby Strings, then reads it
**again after `try_connect` returns** to format the connection-failure message. `try_connect`
blocks and releases the GVL, so another thread can run — and compact — in that window. The
Strings are short (hostnames and socket paths are well under 616 bytes, so their bytes are
embedded in the object slot and relocate with it), and nothing keeps them in place.

Result: on a failed connection, the error message is formatted from freed/moved memory.

## Reproduction

```ruby
require "trilogy"

HOST = +"192.0.2.77"          # TEST-NET-1: unroutable, so connect blocks then fails
$hostref = [HOST]

Thread.new do
  sleep 0.4
  loop do
    GC.verify_compaction_references(expand_heap: true, toward: :empty)
    # size-matched filler: ruby allocates strings from size-pooled heaps, so the
    # filler must match the subject's byte length to reuse the vacated slot
    3.times { 2000.times { +("Z" * HOST.bytesize) } }
    GC.start
  end
end

begin
  Trilogy.new(host: HOST, port: 3306, username: "root",
              connect_timeout: 2, read_timeout: 2, write_timeout: 2)
rescue => e
  puts e.message
end
```

Actual, on trilogy 2.9.0:

```
trilogy_connect - unable to connect to Z:3306
```

Expected (and what you get with the compacting thread removed — I ran that as a control):

```
trilogy_connect - unable to connect to 192.0.2.77:3306
```

The message contains the churn filler, i.e. whatever landed in the vacated slot. Fails 3/3,
control clean 3/3.

## Cause

`contrib/ruby/ext/trilogy-ruby/cext.c` (line numbers from the 2.9.0 gem; HEAD is the same
code at `:668`, `:680`, `:780-782`):

```c
connopt.hostname = StringValueCStr(val);       // :506   raw char* into the Ruby String
...
                                               // `val` is REASSIGNED by every subsequent
                                               // rb_hash_lookup/rb_hash_aref option lookup
...
int rc = try_connect(ctx, &handshake, &connopt);   // :605  blocks, RELEASES THE GVL
if (rc != TRILOGY_OK) {
    if (connopt.path) {
        handle_trilogy_error(ctx, rc, "... %s", connopt.path);        // :608
    } else {
        handle_trilogy_error(ctx, rc, "... %s:%hu", connopt.hostname, // :610
                             connopt.port);
    }
}
```

`trilogy_sock_new` (`src/socket.c:306`) `strdup`s the options immediately, so the
*library-side* copy is fine — the problem is purely the re-read of `connopt.hostname` /
`connopt.path` in the error path. By that point `val` has been overwritten by the later option
lookups, so the original String is not reliably reachable from the machine stack either.

`try_connect` blocks in `_cb_ruby_wait` → `rb_protect(rb_io_wait)`, which releases the GVL, so
this needs no unusual timing — just a concurrent compaction while a connection attempt is
failing.

## Affected versions

| version | result |
| --- | --- |
| 2.9.0 | **FAIL 3/3** — `unable to connect to Z:3306` |
| 2.10.0 | **FAIL 3/3** — `unable to connect to ZZZZZZZZZZ:3306` |
| 2.11.0 | passes — host bytes stayed put |
| 2.12.6 | passes — host bytes stayed put |

The C code at the site is **character-for-character identical across all four**, and unchanged
at HEAD. 2.11.0+ is protected only *incidentally*: `Trilogy#initialize` now creates the socket
in Ruby (`TCPSocket.new(host, port, ...)`, `_connect` arity 3 → 4), which leaves the host
String's `VALUE` on the machine stack for the duration of `_connect`, so conservative stack
scanning pins it. That is a stack-layout side effect of an unrelated refactor, not a fix — the
dangling re-read is still in the code and would resurface if that Ruby-side socket creation
changed.

## Suggested fix

Don't re-read the raw pointers after `try_connect`. Either format the error from the Ruby
`VALUE`s (which are still live and can be kept so with `RB_GC_GUARD`), or capture what the
message needs into a local buffer before the call. `RB_GC_GUARD(val)` alone would not be
enough here, since `val` no longer holds the host String by that point — the guard would have
to be on a dedicated local that keeps the host/socket String referenced across `try_connect`.

## Environment

- trilogy 2.9.0, 2.10.0, 2.11.0, 2.12.6 (gem builds)
- ruby 4.0.6 (2026-07-14 revision 03b6d3f889) +PRISM [arm64-darwin23]
- Sites still present at HEAD: `contrib/ruby/ext/trilogy-ruby/cext.c:668`, `:680`, `:780-782`

byroot:
Fix: https://github.com/trilogy-libraries/trilogy/issues/312

Since we're initializing `connopt` with pointers inside the user provided strings, we must ensure that they are pinned until we return from `rb_trilogy_connect`, otherwise if compaction happens concurrently, or another thread clears the `opts` hash, they could be guarbage collected or moved.

To achieve that, every string option has an associated local variables that is guarded to ensure it isn't optimized out by the compiler.

### DIFF
--- a/contrib/ruby/ext/trilogy-ruby/cext.c
+++ b/contrib/ruby/ext/trilogy-ruby/cext.c
@@ -607,6 +607,18 @@ static int rb_io_descriptor(VALUE io)
 }
 #endif
 
+static inline VALUE config_set_str(VALUE opts, ID key, char **member)
+{
+    VALUE val = rb_hash_aref(opts, ID2SYM(key));
+    if (NIL_P(val)) {
+        return Qfalse;
+    }
+
+    Check_Type(val, T_STRING);
+    *member = StringValueCStr(val);
+    return val;
+}
+
 static VALUE rb_trilogy_connect(VALUE self, VALUE raw_socket, VALUE encoding, VALUE charset, VALUE opts)
 {
     struct trilogy_ctx *ctx = get_ctx(self);
@@ -662,10 +674,10 @@ static VALUE rb_trilogy_connect(VALUE self, VALUE raw_socket, VALUE encoding, VA
         connopt.max_allowed_packet = NUM2SIZET(val);
     }
 
-    if ((val = rb_hash_lookup(opts, ID2SYM(id_host))) != Qnil) {
-        Check_Type(val, T_STRING);
+    VALUE host = config_set_str(opts, id_host, &connopt.hostname);
+    VALUE path;
 
-        connopt.hostname = StringValueCStr(val);
+    if (host) {
         connopt.port = 3306;
 
         if ((val = rb_hash_lookup(opts, ID2SYM(id_port))) != Qnil) {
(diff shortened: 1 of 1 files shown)


contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -791,6 +763,21 @@ static VALUE rb_trilogy_connect(VALUE self, VALUE raw_socket, VALUE encoding, VA
 
     rb_trilogy_release_buffer(ctx);
 
+    // Ensure all our strings are pinned.
+    RB_GC_GUARD(host);
+    RB_GC_GUARD(username);
+    RB_GC_GUARD(password);
+    RB_GC_GUARD(path);
```
jeremy, line 770: Noting `path` may be uninitialized here; not sure whether it matters.

contrib/ruby/ext/trilogy-ruby/cext.c
```diff
@@ -662,10 +674,10 @@ static VALUE rb_trilogy_connect(VALUE self, VALUE raw_socket, VALUE encoding, VA
         connopt.max_allowed_packet = NUM2SIZET(val);
     }
 
-    if ((val = rb_hash_lookup(opts, ID2SYM(id_host))) != Qnil) {
-        Check_Type(val, T_STRING);
+    VALUE host = config_set_str(opts, id_host, &connopt.hostname);
+    VALUE path;
```
jeremy, line 678:
```suggestion
    VALUE path = Qnil;
```

jeremy (commented):
Also noting that `handle_trilogy_error` is `NORETURN` and the guards sit after it… in principle the compiler could treat those VALUEs as dead on a branch that never reaches the guard. Kept live on the clang I checked, but leans on compiler behavior rather than a guarantee.

contrib/ruby/ext/trilogy-ruby/cext.c
byroot: It doesn't.

